PR #10, opened 2017-06-03
Fix iterator error & implements setIn/updateIn

arcanis:
I haven't wrote documentation, and there might be some more work to do to implement move semantic, but I figured you'd be interested to take a look at this. I'm using them, and they seem to work fairly well!

They operate the exact same way than `.set()` and `.update()`, except that they accept any number of key (the compilation will fail if there's too many keys).

### DIFF
--- a/immer/array.hpp
+++ b/immer/array.hpp
@@ -20,8 +20,9 @@
 
 #pragma once
 
-#include <immer/memory_policy.hpp>
 #include <immer/detail/arrays/with_capacity.hpp>
+#include <immer/detail/immutable.hpp>
+#include <immer/memory_policy.hpp>
 
 namespace immer {
 
@@ -236,6 +237,14 @@ public:
     template <typename FnT>
     decltype(auto) update(size_type index, FnT&& fn) &&
     { return update_move(move_t{}, index, std::forward<FnT>(fn)); }
+
+    template <typename ... Rest>
+    auto setIn(size_type index, Rest const & ... rest) const
+    { return detail::setIn(*this, index, rest ...); }
+
+    template <typename ... Rest>
+    auto updateIn(size_type index, Rest const & ... rest) const
+    { return detail::updateIn(*this, index, rest ...); }
 
     /*!
      * Returns a array containing only the first `min(elems, size())`
--- a/immer/detail/immutable.hpp
+++ b/immer/detail/immutable.hpp
@@ -1,0 +1,37 @@
+#pragma once
+
+#include <functional>
+
+namespace immer {
+namespace detail {
+
+template <typename Container, typename Key, typename Updater>
+static inline auto updateIn(Container const & container, Key const & key, Updater const & updater)
+{
+    return container.update(key, updater);
+}
+
+template <typename Container, typename Key, typename ... Rest>
+static inline auto updateIn(Container const & container, Key const & key, Rest const & ... rest)
+{
+    return container.update(key, [&](auto const & child) {
+        return updateIn(child, rest ...);
+    });
+}
+
+template <typename Container, typename Key, typename Value>
+static inline auto setIn(Container const & container, Key const & key, Value const & value)
+{
+    return container.set(key, value);
+}
+
+template <typename Container, typename Key, typename ... Rest>
+static inline auto setIn(Container const & container, Key const & key, Rest const & ... rest)
+{
+    return container.update(key, [&](auto const & child) {
+        return setIn(child, rest ...);
+    });
+}
+
+}
+}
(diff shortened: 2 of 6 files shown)


arximboldi:
Hi @arcanis,

Thanks a lot for your contribution and sorry for the late reply! I am trying to figure a CLA process for the project.

The *first commit* can be merged without problems. If you make a separate pull request for that it can be merged (no CLA needed since changes are minimal). Still, I am curious, how did you find this was missing? Can you add a unit test that shows it? (Not compulsory change)

The *second commit* needs a bit more work, but it really goes in the right direction! I totally want to have those functions there... I am a Clojure user and adding something like this was already in the roadmap, thanks so much for starting the work!

Here are some things that I consider necessary before merging this (I can do them myself if you don't have time, but of course I'd love it if you want to do some of it yourself):

**Style**
- [ ] The library uses the naming conventions of the standard library, this means `updateIn` and `setIn` should be `update_in` and `set_in` respectively.
- [ ] The file `immutable.hpp` is missing the license header (copy one from other file).
- [ ] I normally use `const T&` instead of `T const &`, but that is a minor nitpick and I think it is ok to merge without this change.

**Completeness**
- [ ] Unit tests (in `test/...`).
- [ ] Example (in `example/...`).
- [ ] Documentation (Doxygen style comments).
- [ ] Move semantics. For the _free functions_ this can be achieved by (1) Taking all the arguments as universal references (`T&&` instead of `const T&`), (2) Forwarding the arguments when they are last used (`std::forward<T>(x)`) and (3) changing the return type (`decltype(auto)` instead of `auto`). I can give you some more details on the rationale if you want to. For the member functions, sadly perfect forwarding can not be used for the *this* argument, so you would need to provide two overloads: one that says `const&` after the function signature and is like what you just dhies, and (2) and another one that says `&&` after the function signature and calls `std::move(*this)` when passing control to the free function).

**Design questions**
The point about move semantics shows that there is some boilerplate needed in order to support `set_in` and `update_in` for all containers (basically whatever `vector` does).  Note that I believe this functions are useful for all containers, including `box`! (supporting `box` is a bit trickier though). I consider two possible solutions:

1. Forget about having these functions as members, and just use them as free functions. This is, get `update_in` and `set_in` out of the `detail` namespace and remove the member versions.

2. I concede that for some use-cases the member version feels more natural (specially since other manulators like `update`, `push_back` etc. are members too). This is particularly true for your use-case (multi-dimensional vectors, if I infer properly from the `issue` you opened). An alternative would be to implement the members versions in a mix-in using the CRTP pattern[1] and just include it in all containers.

What do you think about these?

Cheers and thanks a lot!

---
[1] https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Curiously_Recurring_Template_Pattern

*PS.* I am always eager to learn what use-cases people have in mind for these data-structures. If you can talk about it, would you mind telling me a bit about the project you are using these in? 😄

arcanis:
Thanks for the detailed answer, @arximboldi :)

> Still, I am curious, how did you find this was missing? Can you add a unit test that shows it? (Not compulsory change)

Eh eh, I just tried to dereference an iterator using `->` 😛  Trying to call a const member function on a reference obtained this way will break, since the return of the iterator currently is a non-const reference. As for why I was using the `->` operator, I needed to keep track of the previous element while iterating, so a good ol' ranged for wasn't enough.

> 1. Forget about having these functions as members, and just use them as free functions. This is, get update_in and set_in out of the detail namespace and remove the member versions.

From an API standpoint, I think these functions should be member functions, so that they mirror their update / set counterparts.

> 2. An alternative would be to implement the members versions in a mix-in using the CRTP pattern[1] and just include it in all containers.

I personnally tend to avoid mixins. I found that they often make the code much more obscure, adding layers of implicit dependencies. This [blog post](https://facebook.github.io/react/blog/2016/07/13/mixins-considered-harmful.html) talks about React / Javascript, but the philosophy described inside can be useful even for other languages.

> PS. I am always eager to learn what use-cases people have in mind for these data-structures. If you can talk about it, would you mind telling me a bit about the project you are using these in? 😄

I'm writing a small game as a personnal side project, and was looking for a creative way to incrementally alter my GameEnvironment object while still being able to access both the old and the new environment between each game action. Primilarly being a frontend engineer, I was curious to see if using an immutable object would work, since I used them with success in the past.

However, to be completely honest, even if Immer was working pretty fine in term of perfs (mostly because of shared pointers, which avoided a lot of copies - copy on write ftw), I'm in the process of moving back to a mutable architecture, using the STL containers. I think it will eventually give me more control, and I suspect it might work better on more exotic platforms such as mobile. The last nail on the coffin was the C++ ecosystem itself, that makes it really hard to have clean dependencies between projects. Even using a C++ package like Conan seemed like a chore... So I decided to make my core game engine library dependency-free. Of course I won't be able to do the same for the game itself, but at this point I hope I'll be able to just plug my core library into a more mature game framework such as Godot!

Considering that I probably won't have the chance to work more on this PR, I prefer to close it. You're free to use it as inspiration to roll your own variant of update_in and set_in, tho :) And good job on this library, I think it's great to experiment on these sort of things!

PR #26, opened 2017-11-01
Pedantic

prouschal:
I noticed a few warnings when using immer with the compiler options "-Wextra -pedantic".

- extra semicolons
- use of C compound literals for initializing "auto" pointer variables
- unused arguments
- use of extended offsetof for nested subobjects

I fixed the semicolons and compound literals, but not the unused arguments and a use of extended offsetof.

Regarding the compound literals used for initializing pointer variables I opted for a one-line style that preserved the use of auto and uses static_cast. Personally I would have written the type directly, but it seemed out of place among all the auto variables. I am not sure what is preferred here.

In order to test for the changes I turned up the warning level in the tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 project(immer VERSION 0.4.0)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Wno-unused-parameter -Wno-extended-offsetof")
 set(CMAKE_EXPORT_COMPILE_COMMANDS on)
 set(CMAKE_CXX_EXTENSIONS off)
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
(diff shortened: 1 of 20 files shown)


arximboldi:
Thanks! I'll look into it more carefully after MeetingC++

immer/detail/rbts/node.hpp
```diff
@@ -238,7 +238,7 @@ struct node
     {
         assert(n <= branches<B>);
         auto mp = heap::allocate(sizeof_inner_r_n(n));
-        auto mr = (void*){};
+        auto mr = static_cast<void*>(nullptr);
```
arximboldi: Ahá! Good catch! I am wondering whether there might be a more lightweight syntax for AAA pointer declarations... maybe this is ok for now (mutable pointer variables of this kind should not occur often anywayz.)

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 project(immer VERSION 0.4.0)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Wno-unused-parameter -Wno-extended-offsetof")
```
arximboldi: Why did you disable `Wno-extended-offsetof`?

PR #27, opened 2017-11-20
Tny vs2017

tnovotny:
I hope this is what you wanted. This is what I used to run the unit tests. Lemme know if there are any questions / issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,62 +6,67 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
 project(immer VERSION 0.4.0)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
-set(CMAKE_EXPORT_COMPILE_COMMANDS on)
-set(CMAKE_CXX_EXTENSIONS off)
-if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Qunused-arguments")
-endif()
+#set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+#set(CMAKE_EXPORT_COMPILE_COMMANDS on)
+#set(CMAKE_CXX_EXTENSIONS off)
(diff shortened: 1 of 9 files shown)


tnovotny:
This was just a quick way to get VS2017 to accept the CMake file. I use the File|Open|CMake... to open it. I  downloaded and build the gc-7.6.0 and build all libs with CMake and left them in the [x86|x64]-[Release|Debug] dirs. I changed the paths in the CMake file manually. Same goes for the boost stuff. I justed pointed CMake to the location.

tereshkin:
I've tried this request but still got many compilation errors which I fixed manually. Some of them looked like restrictions of c++14 (I wasn't able to use c++17). But also I had many warnings for unused variables and type conversion (from `count_t` to `size_t`).

In the end I managed to get something working for flex_vector<T>. However the numeric result were Ok the unit test system reported some memory leaks.

tnovotny:
What version of the compiler are you using? I tested with 15.5 preview and the 15.4.4.

arximboldi:
Hi! What is the state of this one? Do you think you have time to make it work? Are you using the library on some Vistual Studio projects at the moment?

tnovotny:
I have not made any further changes. I continually retry the version with every build that comes out, both VS 2017 and VS 2017 Preview, but unfort. there are still crashes with the set and map tests (Integer Division by Zero in dada.h). You said that you would look into it, 

> Can you push your code to a branch so I can give it a try?

Which is what I did.

> I have a few hypotheses on what the problem can be but it's probably
fastest if I just try it out myself.

so I was kind of hoping for some feedback from you.

arximboldi:
tnovotny <notifications@github.com> writes:

> I have not made any further changes. I continually retry the version with every build that comes out, both VS 2017 and VS 2017 Preview, but unfort. there are still crashes with the set and map tests (Integer Division by Zero in dada.h). You said that you would look into it, 
>
>> Can you push your code to a branch so I can give it a try?
>
> Which is what I did.
>
>> I have a few hypotheses on what the problem can be but it's probably
> fastest if I just try it out myself.
>
> so I was kind of hoping for some feedback from you.

Hi, thanks for the reminder :)  I'll look into that, sorry it took me so
long to get back to you!

-- 
∿∿∿  https://sinusoid.al
∿∿∿  https://sinusoid.es

arximboldi:
#60 supersedes this PR. Thanks y'all for your contributions!

Issue #31, opened 2018-02-24
Feature Request: Vector constructor from iterator and sentinel


apmccartney:
Firstly, thank you for authoring this library. It's a wonderful contribution to the C++ open source community.

I would propose an additional constructor for `immer::vector` from an **_iterator_** and **_sentinel_**, similar to the existing constructor accepting two iterators. Here, by _**sentinel**_, I mean value of a type equality comparable to an instance of the **_iterator_** type.

My motivation for this request is to improve interoperability with the lazy ranges provided by the [range v3](https://github.com/ericniebler/range-v3) library. In many of these types, the `begin` and `end` methods return values of distinct types. Eric Niebler provides some justification for this peculiarity on his blog [here](http://ericniebler.com/2014/02/21/introducing-iterables/).

arximboldi:
Thanks for the request. I think this is a very sensible request and it should be implemented. We shall me aware of potential ambiguities though, see #30

apmccartney:
If pull requests are welcome, I think I can provide an implementation (along with corresponding tests), with sufficient specificity.

arximboldi:
Austin McCartney <notifications@github.com> writes:

> If pull requests are welcome, I think I can provide an implementation
> (along with corresponding tests), with sufficient specificity.

Of course!

Thanks a lot :)

-- 
∿∿∿  https://sinusoid.al
∿∿∿  https://sinusoid.es

arximboldi:
Thank you very much @apmccartney for addressing this issue!

Issue #30, opened 2018-01-28
immer::vector<int> constructor fails to compile


Toby-Scantlebury:
The constructor `immer::vector<int> aa(1,2);` fails to compile whereas `immer::vector<int> aa(1,2.0);` is fine. It appears that the wrong constructor is being deduced in the first instance.

This is with `gcc version 7.2.0 (Ubuntu 7.2.0-1ubuntu1~16.04)` which reports the following error:

```
In file included from lib/immer/vector.hpp:23:0,
                 from test.cpp:1:
lib/immer/detail/rbts/rbtree.hpp: In instantiation of ‘static auto immer::detail::rbts::rbtree<T, MemoryPolicy, B, BL>::from_range(Iter, Iter) [with Iter = int; T = int; MemoryPolicy = immer::memory_policy<immer::free_list_heap_policy<immer::cpp_heap>, immer::refcount_policy>; unsigned int B = 5; unsigned int BL = 6]’:
lib/immer/vector.hpp:130:35:   required from ‘immer::vector<T, MemoryPolicy, B, BL>::vector(Iter, Iter) [with Iter = int; T = int; MemoryPolicy = immer::memory_policy<immer::free_list_heap_policy<immer::cpp_heap>, immer::refcount_policy>; unsigned int B = 5; unsigned int BL = 6]’
test.cpp:5:26:   required from here
lib/immer/detail/rbts/rbtree.hpp:68:37: error: invalid type argument of unary ‘*’ (have ‘int’)
             result.push_back_mut(e, *first);
```

arximboldi:
Similar problems arise with `std::vector` due to ambiguities in C++'s type
system.  As a workaround, you should be able to dysambiguate using:

    immer::vector<int> aa{{1,2}};

Does that work for you?

Toby-Scantlebury <notifications@github.com> writes:

> The constructor `immer::vector<int> aa(1,2);` fails to compile whereas `immer::vector<int> aa(1,2.0);` is fine. It appears that the wrong constructor is being deduced in the first instance.
>
> This is with `gcc version 7.2.0 (Ubuntu 7.2.0-1ubuntu1~16.04)` which reports the following error:
>
> ```
> In file included from lib/immer/vector.hpp:23:0,
>                  from test.cpp:1:
[...]

PR #32, opened 2018-02-28
added type traits, algorithms, and additional constructors

apmccartney:
Summary of changes: 
The following classes were modified to accomodate iterator sentinel pairs: 
`immer::vector`, `immer::flex_vector`, `immer::array`, `immer::detail::with_capacity`, `immer::detail::no_capacity`, and `immer::detail::node`. To this end, a new file was added, `immer/detail/type_traits.hpp`, providing metaprograms extending the facilities provided by the standard library `type_traits` header, along with several functions to the `immer/algorithm.hpp` header.


apmccartney:
I see there is a test failing due to a memory leak. Was this behavior observed beforehand?

apmccartney:
Closes #31 and #30

apmccartney:
Following up on this pull request. @arximboldi Any idea of when you can review and/or provide feedback?

arximboldi:
@apmccartney sorry for the very late reply, I have been following the PR but haven't had time to look deep and reply yet. I will do a thorough review within one week.

W.r.t. the memory leak, I'm not sure if I have seen this before. I will look into it more carefully later.

arximboldi:
And thank you very much for this effort, it looks very good so far!

immer/algorithm.hpp
```diff
@@ -146,6 +148,23 @@ T accumulate(Iterator first, Iterator last, T init, Fn fn)
...
+          <detail::is_forward_iterator<Iterator>
+           and detail::compatible_sentinel<Iterator,Sentinel>, bool> = true>
+std::size_t distance(Iterator first, Sentinel last){
+    std::size_t result = 0;
+    while(first != last){
+        ++first;
+        ++result;
+    }
+    return result;
+}
```
arximboldi: I know I should fix this with `clang-format` in the long run, but can you try to keep the style consistent with the rest of the code? In particular in this function:
1. New line between function declaration and opening brace of definition.
2. Space between `if`/`while`/`for` keywords and the opening parens after.
3. AAA (Almost Always use Auto)
4. Prefer `&&`/`||` to `and`.

arximboldi: Why is this algorithm added here? For most data-structures there are more efficient ways to compute the distance to a sentinel than just iterating through it. This is particularly true for the data-structures in `immer` so far. I'd rather not provide such a `distance` algorithm or, if provided, provide an optimized version that uses the underlying indexes/structure to be efficient. I can help you if you wanna go in the later direction.

immer/algorithm.hpp
```diff
@@ -221,6 +240,22 @@ bool all_of(Iter first, Iter last, Pred p)
...
+          <detail::compatible_sentinel<SourceIter,Sent>
+           and detail::is_forward_iterator<SinkIter>, bool> = true>
+auto uninitialized_copy(SourceIter first, Sent last, SinkIter d_first){
+    while(first != last){
+        *d_first++ = *first;
+        ++first;
+    }
+    return d_first;
+}
+
```
arximboldi: Same as above: this could be done more efficient in the context of `immer` data-structures.

immer/array.hpp
```diff
@@ -85,23 +85,26 @@ class array
     array() = default;
 
     /*!
-     * Constructs a vector containing the elements in `values`.
+     * Constructs an array containing the elements in `values`.
```
arximboldi: Good catch, thanks for all these typo fixes!

immer/detail/arrays/no_capacity.hpp
```diff
@@ -105,6 +106,19 @@ struct no_capacity
         };
     }
 
+    template <typename Iter, typename Sent,
+              std::enable_if_t
+              <is_forward_iterator<Iter> and
+               compatible_sentinel<Iter, Sent>, bool> = true>
+    static no_capacity from_range(Iter first, Sent last)
+    {
+        auto count = static_cast<size_t>(distance(first, last));
```
arximboldi: Ahá! So I guess you added the `distance` algorithm such that you can use it here, because the `std` one does not support sentinels. Do your sentinels support `last - first` as a way to compute distance? I'd rather have that than the current proposal.

immer/algorithm.hpp
arximboldi: Also, this implementation of `uninitialized_copy` is not correct, because it does not take into account exceptions: if an exception is thrown you have to roll-back, by destructing the objects that have already been copied. This might be the reason why your branch is introducing memory leaks. Still, I would suggest to avoid adding these algorithms here.

immer/detail/type_traits.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+template <typename T, typename = void>
+struct is_dereferenceable_t : std::false_type {};
+
+template <typename T>
+struct is_dereferenceable_t<T, void_t<decltype(*(std::declval<T&>()))>> :
+    std::true_type {};
+
+template <typename T>
+constexpr bool is_dereferenceable = is_dereferenceable_t<T>::value;
```
arximboldi: I would suggest using the same naming conventions as in the `std` for traits:
1. The "naked" trait has no suffix
2. The `::type` alias has `_t` suffix
3. The `::value` or `::type::value` alias has `_v` suffix.

This means you should rename `is_deferenceable_t` as `is_deferenceable`, and rename the current `is_dereferenceable` as `is_dereferenceable_v`.

immer/detail/type_traits.hpp
```diff
@@ -0,0 +1,158 @@
...
+constexpr bool equality_comparable = equality_comparable_t<T, U>::value;
+
+template<typename T, typename U=T, typename = void>
+struct inequality_comparable_t : std::false_type {};
+
+template<typename T, typename U>
+struct inequality_comparable_t
+<T, U, std::enable_if_t
+ <std::is_same
+  <bool, decltype(std::declval<T&>() == std::declval<U&>())>::value>> :
```
arximboldi: Typo: I guess `==` should actually be `!=` here?

test/type_traits/compatible_sentinal.cpp
```diff
@@ -0,0 +1,29 @@
+#include <immer/detail/type_traits.hpp>
```
arximboldi: The filename has a typo `sentinal` -> `sentinel`

arximboldi: You have created lost of test files. I am very happy that you are taking testing seriously! However, I do not think it is necessary to have a different test-file for every different trait. If you ask me, I'd rather merge all these tests in a single file in `test/detail/type_traits.cpp` (also, you did not put the detail in the path of the test files, but for it should be there so it is clearer that these are tests about the internal implementation not the public API.)

immer/detail/type_traits.hpp
```diff
@@ -0,0 +1,158 @@
...
+
+template<typename T, typename U, typename = void>
+struct compatible_sentinel_t : std::false_type {};
+
+template<typename T, typename U>
+struct compatible_sentinel_t
+<T, U, std::enable_if_t
+ <is_iterator<T>
+  and equality_comparable<T, U>
+  and inequality_comparable<T, U>>> :
```
arximboldi: I have mixed feelings about requiring inequality checks for sentinels. What do other people do? In particular, does range v3 require `!=` for sentinels?

apmccartney:
Thank you for the feedback. Over the next day or so, I'll work through to respond or revise. =)

immer/detail/type_traits.hpp
apmccartney: Will do!

immer/detail/type_traits.hpp
apmccartney: Well that's embarassing. Will fix.

test/type_traits/compatible_sentinal.cpp
apmccartney: Will do!

immer/algorithm.hpp
apmccartney:
```
can you try to keep the style consistent with the rest of the code?
```

Will do!

immer/detail/arrays/no_capacity.hpp
apmccartney: > Do your sentinels support last - first as a way to compute distance?

Good point. Iterator/Sentinel pairs do not necessarily support subtraction, but I can check whether this operation is supported and dispatch accordingly.

immer/detail/type_traits.hpp
apmccartney: Whether it is strictly required, I'll have to double-check with one of the range-v3 library authors. I can say with certainty that a defined Iterator/Sentinel inequality operation is assumed in at least some subset of the range-v3 algorithm implementation. See the following [example](https://github.com/ericniebler/range-v3/blob/8ccd974b5cbb91dc9de7ab969abd568fcf569019/include/range/v3/utility/iterator.hpp#L274).

apmccartney:
> There are some nit-picks about formatting and naming...
Good names and consistent formatting matter. Happy to accomodate.

> ... for me the biggest doubt is about the added distance and uninitialized_copy algorithms, which I believe they do not belong in this library.

I understand and agree. Given these overloads will likely be available in the standard library in either the C++20 or C++23 standard library, I felt sheepish implementing them myself.

>  I guess this depends on how you are using sentinels. Are you using sentinels from some particular library? Would it be possible to ADL-select these algorithms from that library instead?

For the moment, the sentinels I have in mind are those provided by range-V3. I must confess that ADL is an area of the language I am woefully unfamiliar with. That said, this presents a good opportunity to educate myself. 

Assuming you aren't in a hurry, I'm happy to experiment with a test project to see whether ADL-selection is possible and share my findings.

> If we conclude that we really need them, I'd rather then maybe move them to detail, prefixed with distance_sentinel uninitialized_copy_sentinel and static assert inside them to make sure they are **only use them when the second type is really a sentinel**.

Emphasis mine. I'm not sure that problem is well formed. =(

A [Sentinel](http://en.cppreference.com/w/cpp/experimental/ranges/iterator/Sentinel)  seem to be defined to be any [semiregular](http://en.cppreference.com/w/cpp/experimental/ranges/concepts/Semiregular) type [equality and inequality comparable](http://en.cppreference.com/w/cpp/experimental/ranges/concepts/WeaklyEqualityComparableWith) to a type fullfilling the [Iterator concept](http://en.cppreference.com/w/cpp/experimental/ranges/iterator/Iterator). In that sense, many iterators are their own sentinels.

> We shall never use them by mistake on a pair of real iterators, since quite possibly the std:: or whatever other library provides is more efficient in that case.

I agree! Would verifying that no standard library implementation exists as part of the dispatch suffice? This has the additional benefit of forward compatibility with C++ 20/23 when the standard library provides overloads for iterator/sentinal pairs.

> Also, in that case, the uninitialized_copy_sentinel has to be improved (now it causes memory leaks when an exception is thrown, see the inline comment for details.)

That was sloppy on my part, and I apologize. Correction incoming.

> Would you like to do these changes yourself or do you need help with that?

With the exception of the ADL question, I feel well within my comfort zone. In either case, I am eager to make the required modification. =)

> This is particularly necessary since Immer is licensed under LGPLv3, which means it can be used in any kind of software (also proprietary), but modifications to the library itself need to be released as LGPLv3 too. This is meant to encourage companies profiting from the library to contribute back. ... I am considering moving to an uber-liberal license like MIT to avoid frictions. I tend to prefer having a minimum of copy-left, specially for software like this that has been a huge time investment for me.

Licensing is an important question and you're right to consider it carefully!

>So I am just gonna ask: what is your take on it, would you be happy to sign a CLA and such that I can keep the the copy-left, or would you rather go for a do-what-evere-you-want kind of license and avoid the CLA?

I also prefer LGPLv3, but understand that not everyone is of that opinion and am pragmatic enough to yield on those topics for the sake of cooperation. 

I am not opposed to submitting to a CLA (conditional on the specifics of the particular agreement).  My request is that such an agreement provide that my contributions will not be subject to an end user license agreement or a proprietary license in the future.

immer/detail/type_traits.hpp
arximboldi: That makes sense, thanks for the link!

immer/detail/type_traits.hpp
arximboldi: No problem, we all make those mistakes :smile:

immer/detail/type_traits.hpp
arximboldi: Thanks! :+1:

immer/detail/arrays/no_capacity.hpp
arximboldi: :+1:

arximboldi:
> For the moment, the sentinels I have in mind are those provided by
> range-V3. I must confess that ADL is an area of the language I am
> woefully unfamiliar with. That said, this presents a good opportunity
> to educate myself.
>
> Assuming you aren't in a hurry, I'm happy to experiment with a test
> project to see whether ADL-selection is possible and share my
> findings.

I am not in a hurry, so feel free to experiment.  Also, fell free to ask
questions here or by email if you get stuck.

>> If we conclude that we really need them, I'd rather then maybe move them to detail, prefixed with distance_sentinel uninitialized_copy_sentinel and static assert inside them to make sure they are **only use them when the second type is really a sentinel**.
>
> Emphasis mine. I'm not sure that problem is well formed. =(
>
> A
> [Sentinel](http://en.cppreference.com/w/cpp/experimental/ranges/iterator/Sentinel)
> seem to be defined to be any
> [semiregular](http://en.cppreference.com/w/cpp/experimental/ranges/concepts/Semiregular)
> type [equality and inequality
> comparable](http://en.cppreference.com/w/cpp/experimental/ranges/concepts/WeaklyEqualityComparableWith)
> to a type fullfilling the [Iterator
> concept](http://en.cppreference.com/w/cpp/experimental/ranges/iterator/Iterator). In
> that sense, many iterators are their own sentinels.

That is correct, a sentinel is more general than an iterator---i.e. an
iterator could be a sentinel, but some sentinels are not iterators.
Furthermore, we are always talking here about sentinels/iterators in the
context of a range, which is either a pair of iterators, or an pair of
an iterator and a sentinel.

Thus, we could argue that if we get passed two things of the same type,
these two things *must* be iterators.  If they are different, then they
may be an iterator/sentinel pair.

>> We shall never use them by mistake on a pair of real iterators, since
>> quite possibly the std:: or whatever other library provides is more
>> efficient in that case.
>
> I agree! Would verifying that no standard library implementation
> exists as part of the dispatch suffice? This has the additional
> benefit of forward compatibility with C++ 20/23 when the standard
> library provides overloads for iterator/sentinal pairs.

That sounds like a good strategy!

>> Also, in that case, the uninitialized_copy_sentinel has to be improved
>> (now it causes memory leaks when an exception is thrown, see the
>> inline comment for details.)
>
> That was sloppy on my part, and I apologize. Correction incoming.

No problem :-)

>> Would you like to do these changes yourself or do you need help with that?
>
> With the exception of the ADL question, I feel well within my comfort
> zone. In either case, I am eager to make the required modification. =)

Thanks a lot for your care and dilligence!  Looking forward to take a
look at your next revision :-_

> I also prefer LGPLv3, but understand that not everyone is of that
> opinion and am pragmatic enough to yield on those topics for the sake
> of cooperation.
>
> I am not opposed to submitting to a CLA (conditional on the specifics
> of the particular agreement).  My request is that such an agreement
> provide that my contributions will not be subject to an end user
> license agreement or a proprietary license in the future.

Thanks for the feedback!  I'll take it into account in my
considerations.

Cheers!

  JP

-- 
∿∿∿  https://sinusoid.al
∿∿∿  https://sinusoid.es

apmccartney:
I've made revisions based on your feedback. 

+ the `distance` and  `uninitialized_copy` algorithm implementations have been constrained such that they will called only when a standard library implementation is unavailable. Otherwise, `immer::distance` and `immer::uninitialized_copy` act as aliases to the standard library implementations.
+ the `type_traits` now follow the standard library convention.
+ `uninitialized_copy` is now exception safe.
+ the `type_traits` tests have been collected in a single file in the `test/detail` directory.
+ new code has been brought with the library style described in the review (I think)

Issues yet to be resolved:

+ the question of ADL look up.
+ the memory leak persists after accounting for exception safety in `uninitialized_copy`. The sanitizer report seems to indicate the leak originates in `__cxxabiv1::__calloc_with_fallback`.

arximboldi:
Austin McCartney <notifications@github.com> writes:

> I've made revisions based on your feedback. 
>
> + the `distance` and  `uninitialized_copy` algorithm implementations have been constrained such that they will called only when a standard library implementation is unavailable. Otherwise, `immer::distance` and `immer::uninitialized_copy` act as aliases to the standard library implementations.
> + the `type_traits` now follow the standard library convention.
> + `uninitialized_copy` is now exception safe.
> + the `type_traits` tests have been collected in a single file in the `test/detail` directory.
> + new code has been brought with the library style described in the review (I think)

Thanks a lot!  I'll take a closer look on monday.

> + the memory leak persists after accounting for exception safety in `uninitialized_copy`. The sanitizer report seems to indicate the leak originates in `__cxxabiv1::__calloc_with_fallback`.

It seems that the issue is also appearing on master, so it might be
related to some build-system vs Travis subtletie.  I will look into that
on monday too!

Cheers!

  JP

immer/algorithm.hpp
```diff
@@ -153,8 +153,8 @@ T accumulate(Iterator first, Iterator last, T init, Fn fn)
  */
 template <typename Iterator, typename Sentinel,
           std::enable_if_t
-	  <detail::std_distance_supports_v<Iterator,Sentinel>, bool> = true>
-std::size_t distance(Iterator first, Sentinel last)
+          <detail::std_distance_supports_v<Iterator,Sentinel>, bool> = true>
+auto distance(Iterator first, Sentinel last)
```
arximboldi: The "Almost Always use Auto" is about variables, not about return types. Sorry if I confused you about that. In this particular case I actually prefer the old version :sweat_smile:

apmccartney: No problem. Changes incoming.

arximboldi:
@apmccartney I just merged the fix for the leaks. If you rebase on master this should go green on Travis...

apmccartney:
doing range-v3 integration testing. Will post results when complete.

apmccartney:
@arximboldi Can you confirm what remains to be addressed is the question of the contirbutor license, or are there additional changes to the submission you'd like to see?

immer/detail/type_traits.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+template <typename T, typename U>
+struct is_subtractable
+<T, U, void_t<decltype(std::declval<T&>() - std::declval<U&>())>> :
+    std::true_type {};
+
+template <typename T, typename U = T>
+constexpr bool is_subtractable_v = is_subtractable<T, U>::value;
+
+using std::swap;
```
arximboldi: I just noticed this. Could you move this line (and the implementation of `is_swapable_with`) to a nested namespace `is_swapable_with_impl` so `std::swap` is not leaked into the `immer::detail` namespace?

apmccartney:
Any developments with the licensing question?

arximboldi:
Austin McCartney <notifications@github.com> writes:

> Any developments with the licensing question?

I will take care of that next week.  I think most issues here are
resolved, I will ping you once that is ready.  Sorry for the long
delays, I hope to be more responsive in future work :)

-- 
∿∿∿  https://sinusoid.al
∿∿∿  https://sinusoid.es

arximboldi:
@apmccartney I have finally decided to just change the license to Boost. I will now proceed to make a commit changing the license and the headers. Once that is ready, you should rebase this branch on top of master. Thanks a lot for your patience!

arximboldi:
The license is changed. @apmccartney can you rebase your changes on master such that they are in agreement with the new license? Also, make sure the right license is used in the header of the newly introduced files. Thank you a lot!

arximboldi:
Hi @apmccartney, how are you? Do you need help with this? Sadly the rebase step is something that you have to do because it is the way you agree to submit the changes under the new license. However if you need any assistance in the process I am happy to help!

apmccartney:
I apologize for the delay and silence. I'll complete the rebase and licensing comments before Sunday. =)

apmccartney:
@arximboldi The fork has been rebased and the license has been updated in files I've added. I've listed the copyright under your name such that future modifcations to the license may be made at your discretion.

arximboldi:
@apmccartney thanks a lot for all your patience and effort! This is a great contribution :) I just did another round of review and will proceed to merge. :tada:

PR #56, opened 2018-07-20
Fix compiler error when iterating a const container

dgel:
Got bitten by a compiler error complaining about conversion from const to non-const here.

### DIFF
--- a/immer/detail/iterator_facade.hpp
+++ b/immer/detail/iterator_facade.hpp
@@ -93,11 +93,11 @@ protected:
     }
 
 public:
-    ReferenceT operator*() const
+    const T& operator*() const
     {
         return access_t::dereference(derived());
     }
-    PointerT operator->() const
+    const T* operator->() const
     {
         return &access_t::dereference(derived());
     }


arximboldi:
Hi @dgel, thanks a lot for your contribution! I don't think the iterator facade is though the right place to fix this issue. Can you please write a unit test that highlights the problem, such that I can find the root cause? If that is too much work, can you just give me an example as a comment?

dgel:
Hi @arximboldi the following snippet will trigger the issue. Sorry if my proposed fix was only a surface fix =)

```cpp
const auto mapping = immer::map<std::string, int>().set("A", 1);
auto it = mapping.begin();
std::cout << it->first << ": " << it->second << std::endl;
```

arximboldi:
Aha! Thanks a lot for the report. I will take a look tomorrow.

arximboldi:
Hi @dgel. A new PR is coming with the proper fix. Also, as a workaround `(*it).first` should work in the meantime.

PR #65, opened 2018-11-24
Map B=6 support.

fedormatantsev:
I still have to refine the implementation, but the idea is clear.
I have some issues with building tests, will continue tomorrow.

Looking forward for the review :)

### DIFF
--- a/immer/detail/hamts/bits.hpp
+++ b/immer/detail/hamts/bits.hpp
@@ -27,13 +27,13 @@ using shift_t  = std::uint32_t;
 template <bits_t B>
 struct get_bitmap_type
 {
-    static_assert(B < 6, "B > 6 is not supported.");
+    static_assert(B < 6u, "B > 6 is not supported.");
 
     using type = std::uint32_t;
 };
 
 template <>
-struct get_bitmap_type<6>
+struct get_bitmap_type<6u>
 {
     using type = std::uint64_t;
 };
@@ -42,13 +42,13 @@ template <bits_t B>
 using bitmap_t = typename get_bitmap_type<B>::type;
 
 template <bits_t B, typename T=count_t>
-constexpr T branches = T{1} << B;
+constexpr T branches = T{1u} << B;
 
 template <bits_t B, typename T=size_t>
-constexpr T mask = branches<B, T> - 1;
+constexpr T mask = branches<B, T> - 1u;
 
 template <bits_t B, typename T=count_t>
-constexpr T max_depth = (sizeof(hash_t) * 8 + B - 1) / B;
+constexpr T max_depth = (sizeof(hash_t) * 8u + B - 1u) / B;
 
 template <bits_t B, typename T=count_t>
 constexpr T max_shift = max_depth<B, count_t> * B;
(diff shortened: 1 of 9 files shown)


immer/detail/hamts/bits.hpp
```diff
@@ -18,43 +18,91 @@ namespace immer {
...
+template <size_t B>
+struct types
+{
+    static_assert(B < 6, "B > 6 is not supported.");
+
+    using bits_t   = std::uint32_t;
+    using bitmap_t = std::uint32_t;
+    using count_t  = std::uint32_t;
+    using shift_t  = std::uint32_t;
+};
```
arximboldi: Hi @fedormatantsev!
One thing that is important: only `bitmap_t` depends on `B`! The other types should remain outside of this `types` template. In this way, you can also keep the `bits_t` type, which you have removed. Also in that way you need to change way less things overall in the code. There are very few mentions to `bitmap_t` in the code, which is the only type dependent on `B`.

immer/detail/hamts/bits.hpp
```diff
@@ -18,43 +18,91 @@ namespace immer {
...
+    return  __popcnt(x);
+#  else
+    return __builtin_popcount(x);
+#  endif
+#else
+    return popcount_fallback(x);
+#endif
+}
+
+inline count_t<6> popcount(bitmap_t<6> x)
```
arximboldi: Instead of overloading popcount for different `bitmap_t<X>`, I would suggest defining the overloads in terms of the underlying representation types. This is, to provide functions:
```
std::uint32_t popcount(std::uint32_t x);
std::uint64_t popcount(std::uint64_t x);
...
```

arximboldi: Otherwise you are unnecesarily coupling a general function (_popcount_) to implementation details of hamts (the notion of _bitmap_). Does this make sense?

immer/detail/hamts/bits.hpp
```diff
@@ -18,43 +18,91 @@ namespace immer {
...
-    return __builtin_popcount(x);
-#  endif
-#else
     // More alternatives:
     // https://en.wikipedia.org/wiki/Hamming_weight
     // http://wm.ite.pl/articles/sse-popcount.html
     // http://graphics.stanford.edu/~seander/bithacks.html#CountBitsSetParallel
     x = x - ((x >> 1) & 0x55555555);
     x = (x & 0x33333333) + ((x >> 2) & 0x33333333);
     return ((x + (x >> 4) & 0xF0F0F0F) * 0x1010101) >> 24;
```
arximboldi: Be careful, this fallback only works for 32 bits, not for any size of bitmap!

immer/detail/hamts/bits.hpp
fedormatantsev: Yeah, sounds reasonable.

immer/detail/hamts/bits.hpp
fedormatantsev: Ah, I will adjust constants based on wiki page
https://en.wikipedia.org/wiki/Hamming_weight
```
int popcount64c(uint64_t x)
{
    x -= (x >> 1) & m1;             //put count of each 2 bits into those 2 bits
    x = (x & m2) + ((x >> 2) & m2); //put count of each 4 bits into those 4 bits 
    x = (x + (x >> 4)) & m4;        //put count of each 8 bits into those 8 bits 
    return (x * h01) >> 56;  //returns left 8 bits of x + (x<<8) + (x<<16) + (x<<24) + ... 
}
```

immer/detail/hamts/bits.hpp
fedormatantsev: Good suggestion, thanks!

fedormatantsev:
At this point I'd like to discuss a special test to check if two values are not colliding when their hash differs only in the last 32 bits. I basically have no idea how to make it, but it actually makes sense to me since tests shouldn't fail even without my changes... I think it requires some kind of introspection for internal data structure

arximboldi:
Hmmm, I don't think that a test needs to instrospect the internal data-structure. I prefer tests that only use the observable interface. There are already tests that generate collisions by overriding the hash function passed. I think it is ok to just reuse the same test-suite and use it for different B's. The tests are already generic, you can look at `immer/test/set/B3.cpp` to see how the tests are instantiated for `B=3`, just create a new file in that folder `B6.cpp` with the appropriate instantiation.

Once again, thanks for doing this work!

fedormatantsev:
I already did B6 tests for map/set =) But the point is not to check if values with hash collision work, but, on the contrary, values with different hashes don't collide

arximboldi:
Nice! This is starting to look very good. It looks to me like it is almost ready, right?

There are a couple of suggestions that I still have. The first one is more stylistic: can you define an alias to the right `using bitmap_t = ...` inside `struct champ` and another one inside `struct node`? In this way, you don't have to tell in all the places `bitmap_t<B>` but just `bitmap_t`, reducing the amount of change but also the amount of places that need to be concerned with dealing with `B`, reducing possibility of mistakes.

The other thing that I don't like a lot is the `get_bit` function. I like to be able to see the bit math directly in the code. I would suggest to just use `bitmap_t{1} << ...` directly in the code instead of going through a function.

fedormatantsev:
Yeah, I think it's ready to be merged.
Some iterations over it might be useful anyway (and I'm ready to iterate(: ).

Feel free to squash&merge the PR whenever you think it's OK.

arximboldi:
This is looking great! I am happy to merge if you want, but I can also wait if you have some other changes in mind. One potential line improvement would be to use 16bit bitmaps for B=4 and 8bit bitmaps for B=3, but it is not a condition for merge. Great work!

fedormatantsev:
I think I'd like to implement your suggestion in scope of another weekend-PR. We can file an issue and assign it to me so I'll not forget to do it.

Let's wait for tests to complete and merge it!

arximboldi:
Done, thanks a lot!

fedormatantsev:
Cool, thanks!

Issue #74, opened 2019-03-06
Segmentation fault with recursive types


jzrake:
I found what appears to be a memory corruption issue with recursive types using `box`. Below is a minimal reproducer which seg-faults on my machine. I am on clang, MacOS 10.14, and `-std=c++14`.

```C++
#include "immer/vector.hpp"
#include "immer/box.hpp"




//=============================================================================
struct my_type
{
    using container_t = immer::vector<immer::box<my_type>>;
    using func_t = std::function<int(int)>;

    int ival;
    double dval;
    func_t func;
    container_t children;
};




//=============================================================================
int main()
{
    my_type::container_t items = {my_type()};
    return 0;
}

```

arximboldi:
Can't reproduce (latest master, both Clang and Gcc on NixOS). Can you please provide a stack trace?

jzrake:
Thanks for the quick reply! Here's some valgrind output -- I hope it helps.


```
Process terminating with default action of signal 11 (SIGSEGV)
==74612==  General Protection Fault
==74612==    at 0x1000049FB: my_type::my_type(my_type&&) (reproducer.cpp:8)
==74612==    by 0x1000049CC: my_type::my_type(my_type&&) (reproducer.cpp:8)
==74612==    by 0x1000049A3: immer::box<my_type, immer::memory_policy<immer::free_list_heap_policy<immer::cpp_heap, 1024ul>, immer::refcount_policy, immer::no_transience_policy, false, true> >::holder::holder<my_type>(my_type&&) (box.hpp:32)
==74612==    by 0x10000494C: immer::box<my_type, immer::memory_policy<immer::free_list_heap_policy<immer::cpp_heap, 1024ul>, immer::refcount_policy, immer::no_transience_policy, false, true> >::holder::holder<my_type>(my_type&&) (box.hpp:32)
==74612==    by 0x1000048A7: immer::box<my_type, immer::memory_policy<immer::free_list_heap_policy<immer::cpp_heap, 1024ul>, immer::refcount_policy, immer::no_transience_policy, false, true> >::holder* immer::detail::make<immer::debug_size_heap<immer::cpp_heap>, immer::box<my_type, immer::memory_policy<immer::free_list_heap_policy<immer::cpp_heap, 1024ul>, immer::refcount_policy, immer::no_transience_policy, false, true> >::holder, my_type>(my_type&&) (util.hpp:64)
[...]

PR #76, opened 2019-03-25
Fix UB on MacOS: use 16 (not 8) byte offset in debug_size_heap

jzrake:
Would resolve issue https://github.com/arximboldi/immer/issues/74

### DIFF
--- a/immer/heap/debug_size_heap.hpp
+++ b/immer/heap/debug_size_heap.hpp
@@ -8,9 +8,10 @@
 
 #pragma once
 
+#include <cassert>
+#include <cstddef>
 #include <immer/config.hpp>
 #include <immer/heap/identity_heap.hpp>
-#include <cassert>
 
 namespace immer {
 
--- a/immer/heap/debug_size_heap.hpp
+++ b/immer/heap/debug_size_heap.hpp
@@ -26,17 +26,17 @@ struct debug_size_heap
     template <typename... Tags>
     static void* allocate(std::size_t size, Tags... tags)
     {
-        auto p = (std::size_t*) Base::allocate(size + sizeof(std::size_t), tags...);
+        auto p = (std::size_t*) Base::allocate(size + 2 * sizeof(std::size_t), tags...);
         *p = size;
-        return p + 1;
+        return p + 2;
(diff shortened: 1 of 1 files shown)


arximboldi:
Hi!

Thanks for submitting the patch and all the investigation involved!

As I explained in my response to #74, the proper fix requires a bit more of work. What I don't like about this PR is that it is not portable and just magically makes the UB disappear for your platform. Since the "proper" fix involves touching the `heap` API it needs a bit more thinking on my side. 

However, you can change your commit to just align to `alignof(std::max_align_t)`. I think the code would look like this (haven't tried it yet):
```cpp
    constexpr static auto extra_size = sizeof(std::aligned_storage<sizeof(std::size_t), alignof(std::max_align_t)>);

    template <typename... Tags>
    static void* allocate(std::size_t size, Tags... tags)
    {
        auto p = (std::size_t*) Base::allocate(size + extra_size, tags...);
        *p = size;
        return ((char*)p) + extra_size;
    }

    template <typename... Tags>
    static void deallocate(std::size_t size, void* data, Tags... tags)
    {
        auto p = (std::size_t*) (((char*) data) - extra_size);
        assert(*p == size);
        Base::deallocate(size + extra_size, p, tags...);
    }
```
Can you give it a try?

arximboldi:
Also, note that the `free_list` heap family is working just because we are lucky... I can fix those later myself if you just want to merge this sooner so you can merge progress.

jzrake:
Ok - there's an update to the PR.

Let me know if/how I can help in resolving similar issues to the `free_lists`'s.

arximboldi:
Did this version fix it on your machine?

arximboldi:
(It seems there are issues on Travis ;) )

jzrake:
Oh no! Yes, (my own) basic examples work on my machine as of c106d27426f47aad47f2a7a3e24b78e9b7fda92b. How can I run the full test suite locally?

arximboldi:
Seems to be working now. Thanks a lot for working through PR!
I hope you enjoy using Immer, looking forward to hear more about what you build with it :smile:

PR #77, opened 2019-04-01
Only use IMMER_SMT_PAUSE when it is defined

codablock:
IMMER_SMT_PAUSE won't be defined on ARM builds, so we need to fall back to
using std::this_thread::yield().


arximboldi:
Good catch. Thanks!

Issue #78, opened 2019-04-01
Assertion failure on mingw 32bit builds


codablock:
https://github.com/arximboldi/immer/pull/76 introduced the use of `alignof(std::max_align_t)` to determine the offset of the `size` value inside the allocated memory.

There is however an issue when built with mingw and 32bits. The bug described in https://sourceforge.net/p/mingw-w64/bugs/778/ results in `extra_size` being different, depending on the order of inclusion of immer headers and standard headers.

In my specific case, this results in allocation happening with `extra_size` being 16 and deallocation happening with `extra_size` being 8, which then causes the assertion in `debug_size_heap::deallocate` to fail.

I was considering creating a PR, but then was not sure how to fix this so that it would be accepted. Possible solution would be to fix `extra_size` to always be 8 bytes (e.g. `2 * sizeof(void*)`) when mingw+32bit is detected. I don't see how this can be fixed through the inclusion order, as inside immer you have no control about what was included before the first immer header.

arximboldi:
I am happy to accept a fix that looks like:
```
#if defined(__MINGW32__) && !defined(__MINGW64__)
// workaround
#else 
// normal flow
#endif
```
Note that the check is needed in that form because mingw32 defines the first macro also for 64 bit builds. I don't have a MING32 system at hand, so I'd thank you if you could make such PR and try it yourself.

Thanks a lot!

arximboldi:
Wait a second... I just realized that the default case is also incorrect. I thought I accepted something else in the last PR touching that code, dunno what happened in between...

codablock:
Ok, I'll create a PR after your latest PR which touches the same line is merged.

arximboldi:
Should be fixed now with your PR. Thanks a lot Alex for the bugreport and fix!

jzrake:
Oops, that was my mistake in #76, sorry about that. Arx had suggested the version with `std::aligned_storage_t` in the first place; I thought (wrongly) they're always equivalent.

codablock:
@jzrake As I understand the issue, it was not about wether using `aligned_storage_t` or not. It's the use of `std::max_align_t`, independent of how you use it, that caused the issue on MinGW 32bit. Nothing you could have expected to happen, so all good :

arximboldi:
Don't worry @jzrake, the `aligned_storage_t` thing only fixes the situation where `sizeof(size_t)` is smaller than the max alignment, which to be honest I am not sure there is any architecture alive where it's the case.

PR #80, opened 2019-04-01
Fix debug_size_heap::extra_size to always be 8 on MinGW 32bit builds

codablock:
As discussed in https://github.com/arximboldi/immer/issues/78

I chose to hardcode it to 8 because the alternative of `sizeof(void*) * 2` is not really connected to how `std::max_align_t` is defined.


arximboldi:
Thanks a lot! :+1:

PR #85, opened 2019-06-07
Add missing include call

myd7349:
`CMAKE_INSTALL_INCLUDEDIR` and `CMAKE_INSTALL_LIBDIR` used in `CMakeLists.txt` are defined in `GNUInstallDirs.cmake`. Without including this module, this command will fail:
```bash
$ cmake --build . --target install
```
>  -- Install configuration: "Debug"
  -- Up-to-date: /cmake/Immer/ImmerConfig.cmake
  CMake Error at cmake_install.cmake:60 (file):
    file INSTALL given no DESTINATION


arximboldi:
Thanks @myd7349 !

PR #86, opened 2019-06-11
Fix warnings with msvc + [[nodiscard]] support

mrpi:
The macro to check for "[[nodiscard]]" compiler support did not work with MSVC 19.16 and produced warnings:
https://godbolt.org/z/2-EjTj
"warning C4067: unexpected tokens following preprocessor directive - expected a newline"

This patch removes the warning and enables the attribute, if the Microsoft compiler is running in C++17 mode.

### DIFF
--- a/immer/config.hpp
+++ b/immer/config.hpp
@@ -8,14 +8,17 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


arximboldi:
Good catch, thanks a lot @mrpi !

PR #94, opened 2019-07-05
Fix issues with unaligned data types on immer::set

arximboldi:
RFC @mrpi!

This fixes issue #88. It just makes the test you gave me pass, but I have not checked with other data-structures. Did you face similar problems with other containers indeed?

### DIFF
--- a/immer/detail/hamts/node.hpp
+++ b/immer/detail/hamts/node.hpp
@@ -92,8 +92,9 @@ struct node
 
     constexpr static std::size_t sizeof_values_n(count_t count)
     {
-        return immer_offsetof(values_t, d.buffer)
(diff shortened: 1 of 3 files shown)


mrpi:
Thank you very much!

> Did you face similar problems with other containers indeed?

No, my problem occurred with a _struct_ whose only data member was a _std::array<char, 10>_ put into an _immer::set_.
This triggered the error on 32bit, I had to enlarge the std::array to reproduce the bug on 64bit.
I have not tried to put this _struct_ into other containers from your library. Neither have i tried classes/structs with other "exotic" alignment requirements.

arximboldi:
Ok, thanks! I will merge this for now and do a bit of experimentation myself when I have time.

PR #98, opened 2019-09-05
Do not set value of IMMER_{HAMTS,RBTS}_TAGGED_NODE if it is already set

dwightguth:
Currently the data layout of the immer data types depends on the value
of the NDEBUG macro. This is rather unfortunate because if a library is
built with debugging disabled, and then linked into an application built
with debugging enabled, the code will crash with an assertion error.

Thus, I provide a way to explicitly set the layout intended for the
application so that it does not get overridden by the value of NDEBUG,
which allows the user to have a consistent data layout across multiple
code units with different values of debugging.

### DIFF
--- a/immer/detail/hamts/node.hpp
+++ b/immer/detail/hamts/node.hpp
@@ -14,10 +14,12 @@
 
 #include <cassert>
 
+#ifndef IMMER_HAMTS_TAGGED_NODE
 #ifdef NDEBUG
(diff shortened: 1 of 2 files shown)


dwightguth:
This is proving more complicated than I expected to get working, so I am going to close this PR and recreate it when I get it working.

arximboldi:
Cool. Thanks for your other PR!

PR #99, opened 2019-09-05
Make tagged nodes configurable

dwightguth:
It's unfortunate that the data layout of the library depends on the value of the NDEBUG macro, because it makes the code throw an assertion error if a library that includes immer is built with NDEBUG and then an application that includes immer but does not set NDEBUG links against that library. So I moved the functionality to set whether tagged nodes exist into config.hpp, and allowed it to be set by the user. We then create a new macro, assertKind, which we use instead of assert to guard the assertions that depend on tagged nodes.

### DIFF
--- a/immer/config.hpp
+++ b/immer/config.hpp
@@ -22,6 +22,20 @@
 #define IMMER_NODISCARD
 #endif
 
+#ifndef IMMER_TAGGED_NODE
+ #ifdef NDEBUG
+  #define IMMER_TAGGED_NODE 0
+ #else
+  #define IMMER_TAGGED_NODE 1
+ #endif
+#endif
+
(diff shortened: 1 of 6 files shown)


arximboldi:
Hi @dwightguth,

Thanks a lot for this change. I think it is indeed quite useful.

My only concern is that `assertKind()` is (1) not namespaced, and (2) does not follow the naming conventions. Would you mind changing it to `IMMER_ASSERT_TAGGED`?

Otherwise I'm happy to merge!

dwightguth:
@arximboldi I made the rename you suggested. Let me know if you have any other suggestions.

arximboldi:
That was very responsive, thanks for making the change! Merging now.

PR #109, opened 2019-10-04
add fuzzers for array, map and set

luismerino:
Some work in progress, just for the shake of sharing the code

### DIFF
--- a/extra/fuzzer/array.cpp
+++ b/extra/fuzzer/array.cpp
@@ -1,0 +1,82 @@
+//
+// immer: immutable data structures for C++
+// Copyright (C) 2016, 2017, 2018 Juan Pedro Bolivar Puente
+//
+// This software is distributed under the Boost Software License, Version 1.0.
+// See accompanying file LICENSE or copy at http://boost.org/LICENSE_1_0.txt
+//
+
+#include "fuzzer_input.hpp"
+#include <immer/array.hpp>
+#include <iostream>
+#include <array>
+
+extern "C"
+int LLVMFuzzerTestOneInput(const std::uint8_t* data, std::size_t size)
+{
+    constexpr auto var_count = 4;
+
+    using array_t= immer::array<int, immer::default_memory_policy>;
+    using size_t   = std::uint8_t;
+
+    auto vars = std::array<array_t, var_count>{};
+
+    auto is_valid_var = [&] (auto idx) {
+        return idx >= 0 && idx < var_count;
+    };
+    auto is_valid_index = [] (auto& v) {
+        return [&] (auto idx) { return idx >= 0 && idx < v.size(); };
+    };
+    auto is_valid_size = [] (auto& v) {
+        return [&] (auto idx) { return idx >= 0 && idx <= v.size(); };
+    };
+
+    return fuzzer_input{data, size}.run([&] (auto& in)
+    {
+        enum ops {
+            op_push_back,
+            op_update,
+            op_take,
+            op_push_back_move,
+            op_update_move,
+            op_take_move,
+        };
+        auto src = read<char>(in, is_valid_var);
+        auto dst = read<char>(in, is_valid_var);
+        switch (read<char>(in))
+        {
+        case op_push_back: {
+            vars[dst] = vars[src].push_back(42);
+            break;
+        }
+        case op_update: {
+            auto idx = read<size_t>(in, is_valid_index(vars[src]));
+            vars[dst] = vars[src].update(idx, [] (auto x) { return x + 1; });
+            break;
+        }
+        case op_take: {
+            auto idx = read<size_t>(in, is_valid_size(vars[src]));
+            vars[dst] = vars[src].take(idx);
+            break;
+        }
+        case op_push_back_move: {
+            vars[dst] = std::move(vars[src]).push_back(12);
+            break;
+        }
+        case op_update_move: {
+            auto idx = read<size_t>(in, is_valid_index(vars[src]));
+            vars[dst] = std::move(vars[src])
+                .update(idx, [] (auto x) { return x + 1; });
+            break;
+        }
+        case op_take_move: {
+            auto idx = read<size_t>(in, is_valid_size(vars[src]));
+            vars[dst] = std::move(vars[src]).take(idx);
+            break;
+        }
+        default:
+            break;
+        };
+        return true;
+    });
+}
--- a/extra/fuzzer/map-gc.cpp
+++ b/extra/fuzzer/map-gc.cpp
@@ -25,21 +25,12 @@ int LLVMFuzzerTestOneInput(const std::uint8_t* data, std::size_t size)
     constexpr auto var_count = 4;
 
     using map_t = immer::map<char, int, std::hash<char>, std::equal_to<char>, gc_memory>;
-    // using size_t   = std::uint8_t;
 
     auto vars = std::array<map_t, var_count>{};
 
     auto is_valid_var = [&] (auto idx) {
         return idx >= 0 && idx < var_count;
     };
-    /*
-    auto is_valid_index = [] (auto& v) {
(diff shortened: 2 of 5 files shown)


extra/fuzzer/map-gc.cpp
```diff
@@ -0,0 +1,105 @@
...
+    immer::gc_transience_policy,
+    false>;
+
+extern "C"
+int LLVMFuzzerTestOneInput(const std::uint8_t* data, std::size_t size)
+{
+    constexpr auto var_count = 4;
+
+    using map_t = immer::map<char, int, std::hash<char>, std::equal_to<char>, gc_memory>;
+    // using size_t   = std::uint8_t;
```
arximboldi: Please remove commented code.

extra/fuzzer/map-gc.cpp
```diff
@@ -0,0 +1,105 @@
...
+        return idx >= 0 && idx < var_count;
+    };
+    /*
+    auto is_valid_index = [] (auto& v) {
+        return [&] (auto idx) { return idx >= 0 && idx < v.size(); };
+    };
+    auto is_valid_size = [] (auto& v) {
+        return [&] (auto idx) { return idx >= 0 && idx <= v.size(); };
+    };
+    */
```
arximboldi: Same.

extra/fuzzer/map-gc.cpp
```diff
@@ -0,0 +1,105 @@
...
+    return fuzzer_input{data, size}.run([&] (auto& in)
+    {
+        enum ops {
+            op_set,
+            op_erase,
+            op_set_move,
+            op_erase_move,
+            iterate,
+            find,
+            update
```
arximboldi: Mising `op_` prefix.

extra/fuzzer/map-gc.cpp
```diff
@@ -0,0 +1,105 @@
...
+        case op_set_move: {
+            auto value = read<size_t>(in);
+            vars[dst] = std::move(vars[src]).set(value, 42);
+            break;
+        }
+        case op_erase_move: {
+            auto value = read<size_t>(in);
+            vars[dst] = std::move(vars[src]).erase(value);
+            break;
+        }
```
arximboldi: Note: we have not implemented transients for set/erase yet, so these are actually identical to the non move versions for now. It still makes sense to have them here though.

extra/fuzzer/map-gc.cpp
```diff
@@ -0,0 +1,105 @@
...
+        case iterate: {
+            if(src != dst) {
+                for(auto it = vars[src].begin(); it != vars[src].end(); ++it) {
+                    vars[dst] = vars[dst].set(it->first, it->second);
+                }
+            }
```
arximboldi: Instead of doing the `src != dst` check, we can otherwise just make sure the original `src` value is kept alive during the operation. Also, we can use a C++11 for each here. All this considered, this may be rewritten as:
```c++
auto srcv = vars[src];
for (const auto& v : srv) {
    vars[dst] = vars[dst].set(v.first, v.second);
}
```
On the other hand, beyond iteration, this operation is mergin `src` into `dst`. Is this what we want, or do we want to focus this operation on iteration alone? I guess the merging ensures the iteration does not get optimized away, but there are maybe more lightweight ways to do this? This would mean though, that `dst` does not affect the operation, so maybe we should only read `dst` for binary operations... [This would also be applied to the fuzzers of the vectors. Also, another digression, maybe it makes sense to pack both src and dst in one byte of the input. Maybe we can leave this like this for now and think about it later...]

extra/fuzzer/map-gc.cpp
```diff
@@ -0,0 +1,105 @@
...
+            break;
+        }
+        case find: {
+            auto value = read<size_t>(in);
+            auto res = vars[src].find(value);
+            if(res != nullptr) {
+                vars[dst] = vars[dst].set(*res, 42);
```
arximboldi: [Same digression about using `dst` in "read only" operations.]

extra/fuzzer/map.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include <iostream>
+#include <array>
+
+extern "C"
+int LLVMFuzzerTestOneInput(const std::uint8_t* data, std::size_t size)
+{
+    constexpr auto var_count = 4;
+
+    using map_t = immer::map<char, int>;
+    // using size_t   = std::uint8_t;
```
arximboldi: Same.

extra/fuzzer/map.cpp
```diff
@@ -0,0 +1,97 @@
...
+        return idx >= 0 && idx < var_count;
+    };
+    /*
+    auto is_valid_index = [] (auto& v) {
+        return [&] (auto idx) { return idx >= 0 && idx < v.size(); };
+    };
+    auto is_valid_size = [] (auto& v) {
+        return [&] (auto idx) { return idx >= 0 && idx <= v.size(); };
+    };
+    */
```
arximboldi: Tralari.

extra/fuzzer/map.cpp
```diff
@@ -0,0 +1,97 @@
...
+    return fuzzer_input{data, size}.run([&] (auto& in)
+    {
+        enum ops {
+            op_set,
+            op_erase,
+            op_set_move,
+            op_erase_move,
+            iterate,
+            find,
+            update
```
arximboldi: `op_` :)

extra/fuzzer/map.cpp
```diff
@@ -0,0 +1,97 @@
...
+            break;
+        }
+        case find: {
+            auto value = read<size_t>(in);
+            auto res = vars[src].find(value);
+            if(res != nullptr) {
+                vars[dst] = vars[dst].set(*res, 42);
+            }
+            break;
+        }
```
arximboldi: Same digression.

extra/fuzzer/set-gc.cpp
```diff
@@ -0,0 +1,90 @@
+//
```
arximboldi: I'm wondering if, now that we are gonna treat Fuzzers more seriously and maintain them actively, we shall for example do the same kind of tricks we do in the unit tests to reuse the same testing code for all the variations of the `MemoryPolicy` (gc, non gc, etc.)

extra/fuzzer/set-gc.cpp
```diff
@@ -0,0 +1,90 @@
...
+                }
+            }
+            break;
+        }
+        default:
+            break;
+        };
+        return true;
+    });
+}
```
arximboldi: A lot of stuff same as for `map/`

extra/fuzzer/set.cpp
```diff
@@ -0,0 +1,82 @@
...
+                }
+            }
+            break;
+        }
+        default:
+            break;
+        };
+        return true;
+    });
+}
```
arximboldi: Same as for `map`.

extra/fuzzer/array.cpp
```diff
@@ -0,0 +1,82 @@
+//
```
arximboldi: Maybe we should reuse the code with `vector`, like we do for unit tests.

extra/fuzzer/map-gc.cpp
luismerino: I like merging here, I guess it gives more chances of triggering errors than just iterating :)

luismerino:
Thanks for your quick comments! I have pushed some of the requested changes. What do you think?

extra/fuzzer/map-gc.cpp
arximboldi: Not sure about that logic. Having operations more "orthogonal" gives the fuzzer more options to combine operations in surprising ways.

arximboldi:
Thanks for the work @luismerino! Yeah I think we can merge this version and iterate later.

Issue #11, opened 2017-06-13
Add support for VS 2017


acki-m:
I tried compilation of the latest master with VS2017,
but I get a compile error:
```
inline constexpr auto clz_(unsigned int x) { return __builtin_clz(x); }
immer/detail/util.hpp(82): error C3861: '__builtin_clz': identifier not found
```
Can you please add support for VS 2017?

arximboldi:
Hi @acki-m!

Thanks a lot for reporting the issue. I agree that this is an important feature. In it somewhere in the roadmap, but at the moment I must admit there are other features above that.

If this is important for a commercial project you are working on, I suggest joining the sponsorship program. Alternatively we could agree on some consulting time for your project to help finance this work. In this way I could guarantee to develop it in a timeline tailored to your needs, and I could also do extra work to make sure the library works for you. Send me an to juanpe@sinusoid.al if you are interested!

PS. I see you are in Berlin... I live there too, so we could meet in offline space to discuss this further too. I also attend the C++ meetup often.

acki-m:
Hi Juan,
We actually meet each other, I attended the C++ meet up where you presented your library. Which was actually one of the best presentation I've heard :thumbsup:. And that lead me to your github page.

At the moment I just want to evaluate the library for myself and cannot sponsor anything.
What I can sponsor you is setting up AppVeyor-CI myself to test your library under windows.
However, I can only recommend you to invest supporting the MSVC-Toolchain. There are a lot of potential customers.

arximboldi:
Thanks a lot for the kind words 😊 

I'll keep the issue open and ping you back when I get started with the MSVC work so you can test it.

tnovotny:
i can get this to compile with VS2017 using /std:c++17

    #include "immer/flex_vector.hpp"
    
    int main() {
        immer::flex_vector<int> is;
    }

by patching the log2 calculation (line 87)

    #if 0

    template <typename T>
    inline constexpr auto clz_(T) -> not_supported_t { IMMER_UNREACHABLE; return {}; }
    inline constexpr auto clz_(unsigned int x) { return __builtin_clz(x); }
    inline constexpr auto clz_(unsigned long x) { return __builtin_clzl(x); }
    inline constexpr auto clz_(unsigned long long x) { return __builtin_clzll(x); }

    template <typename T>
    inline constexpr T log2_aux(T x, T r = 0)
    {
        return x <= 1 ? r : log2(x >> 1, r + 1);
    }

    template <typename T>
[...]

PR #113, opened 2019-11-25
Fix position.hpp integer shift warnings on 64-bit systems

jimbo1qaz:
This replaces all occurrences of `1 << shift_` with `size_t{1} << shift_`.

Fixes "[potential] issues in containers with more than 2^32 elements (on amd64 machines at least)." See #11.

### DIFF
--- a/immer/detail/rbts/position.hpp
+++ b/immer/detail/rbts/position.hpp
@@ -740,7 +740,7 @@ struct regular_sub_pos
     {
         return offset == subindex(size_ - 1)
             ? size_ - size_before(offset)
-            : 1 << shift_;
+            : size_t{1} << shift_;
     }
 
     auto size_sbh(count_t offset, size_t size_before_hint)
@@ -748,7 +748,7 @@ struct regular_sub_pos
         assert(size_before_hint == size_before(offset));
         return offset == subindex(size_ - 1)
             ? size_ - size_before_hint
-            : 1 << shift_;
+            : size_t{1} << shift_;
     }
 
     void copy_sizes(count_t offset,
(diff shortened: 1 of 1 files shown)


jimbo1qaz:
Should I add a commit with MSVC-specific `#pragma` to disable warnings mentioned in #11?

arximboldi:
Looks great, thanks! Which warnings are you referring to btw, I thought this commit fixes the warnings?

jimbo1qaz:
https://github.com/arximboldi/immer/issues/11#issuecomment-508731374

> `immer\detail\rbts\rrbtree.hpp(956): warning C4267: 'argument': conversion from 'size_t' to 'immer::detail::rbts::count_t', possible loss of data`

to which you replied: https://github.com/arximboldi/immer/issues/11#issuecomment-508753298

> The first warning is actually ignorable, the code is correct because `tail_size == branches<BL>`, which is a small and in the `count_t` domain. Actually probably the problem goes away by replacing `tail_size` by `braches<BL>` in that branch.

Some of the warnings: https://gist.github.com/jimbo1qaz/f13deda23c2f59cf227fc5bdd4627765
rrbtree.hpp for line number reference: https://gist.github.com/jimbo1qaz/f13deda23c2f59cf227fc5bdd4627765#file-rrbtree-hpp

arximboldi:
Oh, thanks! Is there a more cross-platform way of getting rid of that warning? E.g. using a `static_cast` where the implicit conversion is now ocurring.

jimbo1qaz:
Maybe merge this (possible bug-fix) first, and I'll report that in a separate issue?

Looks like there's a lot of places this cast would need to be added. In my original full compiler log, I found 31 unique line numbers affected by C4267: https://gist.github.com/jimbo1qaz/f13deda23c2f59cf227fc5bdd4627765#file-immer-msvc-warnings-c4267-unique

arximboldi:
Uhhh, damn, yeah... Maybe then we could just go with a pragma over the whole involved header with a comment. I'm gonna merge this feel free to make another PR with the pragma for that error.

Issue #118, opened 2020-01-02
Fix warning in Xcode 11


zeyangl:
immer now generates the following warnings in Xcode 11(and similar in Visual Studio 2019) 
![image](https://user-images.githubusercontent.com/1244606/71654618-aae45f80-2d6d-11ea-816f-95fe587cbbb0.png)
It would be nice if these could be fixed.

arximboldi:
Thanks, I will look into fixing them. Which `clang` version and warning flags do you have enabled?

arximboldi:
(In any case, the warnings are innocuous, the conversions are correct since the values are whithin range.)

zeyangl:
Yes they are innocuous indeed, just a bit annoying. If the fix is trivial and harmless then probably better to have it than not. I believe the -Wconversion is enabled by default in Xcode's clang(likely 8.0) but not so with the nix-shell env.

arximboldi:
Yes. Sadly the fix suggested is not harmless since it changes the sizes of things. I will look into another fix. Thanks anyways for your efforts!

arximboldi:
Closing this for now. You should anyway include external dependencies with `-isystem` to avoid such false positives.

PR #123, opened 2020-03-01
fix a few type conversion warnings

zeyangl:
This is to address the issue https://github.com/arximboldi/immer/issues/118
All tests with make check passed. Ran with clang-7.1


zeyangl:
it seems 2 of the 7 checks failed due to a network error when pulling the build environemnt.

arximboldi:
`count_t` does not need to be `size_t`. Those warnings are innocuous.

PR #124, opened 2020-03-05
Minimal changes to support inclusion via add_subdirectory()

audiofanatic:
When immer is used as part of a larger project hierarchy, the parent
will usually want/need to disable the non-essential parts of immer.
Some of the tests, examples, etc. will clash with things from other
projects, so being able to disable them in a project-specific way is
essential.


arximboldi:
Nice! Thanks! :

PR #125, opened 2020-03-29
Minor fixes: MSVC static analysis warnings

ikrima:
- Add defined(__GNUC__) guard to include to avoid "undefined symbol" warning
  - iterator_facade doesn't need to derive from std::iterator; should explicitly define type categories instead

### DIFF
--- a/immer/detail/iterator_facade.hpp
+++ b/immer/detail/iterator_facade.hpp
@@ -52,12 +52,12 @@ template <typename DerivedT,
           typename DifferenceTypeT = std::ptrdiff_t,
           typename PointerT = T*>
 class iterator_facade
-    : public std::iterator<IteratorCategoryT,
-                           T,
-                           DifferenceTypeT,
-                           PointerT,
-                           ReferenceT>
 {
+  using iterator_category = IteratorCategoryT;
+  using value_type        = T;
+  using difference_type   = DifferenceTypeT;
+  using pointer           = PointerT;
+  using reference         = ReferenceT;
 protected:
     using access_t = iterator_core_access;
 
--- a/immer/detail/combine_standard_layout.hpp
+++ b/immer/detail/combine_standard_layout.hpp
@@ -10,7 +10,7 @@
 
 #include <type_traits>
 
-#if __GNUC__ == 7 && __GNUC_MINOR__ == 1
+#if defined(__GNUC__) && __GNUC__ == 7 && __GNUC_MINOR__ == 1
 #define IMMER_BROKEN_STANDARD_LAYOUT_DETECTION 1
 #define immer_offsetof(st, m) ((std::size_t) &(((st*)0)->m))
 #else


immer/detail/iterator_facade.hpp
```diff
@@ -52,12 +52,12 @@ template <typename DerivedT,
           typename DifferenceTypeT = std::ptrdiff_t,
           typename PointerT = T*>
 class iterator_facade
-    : public std::iterator<IteratorCategoryT,
-                           T,
-                           DifferenceTypeT,
-                           PointerT,
-                           ReferenceT>
```
arximboldi, line 59: What is the purpose of this change?

ikrima: **High order bit:** I'd defer to you on good/bad on that. Don't have any opinions/was only cleaning code/l

Motivation was purely getting rid of MSVC deprecation warnings (MSVC 2019, 16.5.1).
In the middle of a different branch/repo, but IIRC, this was the specific warning:

> D.13 [depr.iterator.primitives]: “warning STL4015: The std::iterator class template (used as a base class to provide typedefs) is deprecated in C++17. (The <iterator> header is NOT deprecated.) The C++ Standard has never required user-defined iterators to derive from std::iterator. To fix this warning, stop deriving from std::iterator and start providing publicly accessible typedefs named iterator_category, value_type, difference_type, pointer, and reference. Note that value_type is required to be non-const, even for constant iterators. You can define _SILENCE_CXX17_ITERATOR_BASE_CLASS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to acknowledge that you have received this warning.”

[This https://devblogs.microsoft.com/cppblog/c17-feature-removals-and-deprecations/](This https://devblogs.microsoft.com/cppblog/c17-feature-removals-and-deprecations/)

Other option is to #define  _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS  for MSVC targets

arximboldi: I agree with your change then, thank you!

audiofanatic:
Sadly this seemingly reasonable change can break VS2019 builds (version tested is 16.4.6). I can't provide a minimal example right now, but I have code that breaks with a structure something like the following. Hopefully it is enough to be able to replicate it:

```c++
// src is of type immer::map<int,T>
std::vector<std::pair<int,T>> dest(src.size());
std::copy(src.begin(), src.end(), dest.begin());
```

The error message looks something like this (I've had to replace the real types, but it should match up with the reduced example above):

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.24.28314\include\xutility(1018): error C2794: 'iterator_category': is not a member of any direct or indirect base class of 'std::iterator_traits<_Iter>'
        with
        [
            _Iter=immer::detail::hamts::champ_iterator<std::pair<int,T>,immer::map<int,T,std::hash<int>,std::equal_to<int>,immer::default_memory_policy,5>::hash_key,immer::map<int,T,std::hash<int>,std::equal_to<int>,immer::default_memory_policy,5>::equal_key,immer::default_memory_policy,5>
        ]
```

Ignoring whether the code is optimal or not, it is nonetheless valid. There's an easy enough workaround (e.g. a manual for loop), but it would be great to not have to resort to that.

arximboldi:
I just noticed what the problem is, I truly merged this PR carelessly, the new aliases are introduced as private! Also the code is not even indented with four spaces... :p Fixing it now! Thanks for bringing this up @audiofanatic !

ikrima:
oof. hello there my old friend, cone of shame; it's been a while since we hung out :P

mae culpa is mine. Carelessly just cut & pasted the code segment from my local repo instead of forking a separate branch bc "it's straightforward" *smh*

Apologies!

arximboldi:
No prob @ikrima, thanks anyways for finding the original issue and being proactive about solving it!

PR #132, opened 2020-05-25
Added set::find

### LINKED ISSUE #121: Missing immer::set<T>::find( ) method.
Hi, currently it seems that immer::set provides count( ) but not find( ), while immer::map provides both.  Would it be possible to add an immer::set<T>::find( ) function?

I have a data type that has some fields that are used as annotations.  They affect neither equality (==) or the hash function.  I am interested in looking up an element, and finding the (possibly different) version of that element in the set, and check to see if their notes are equal.  This is possible with std::set, and it is possible with immer, but only if you walk through the entire set.

One might argue that this is bad design, but it DOES make transitioning from std::set< > more difficult.

If you are opposed to this, maybe a note in the header could be added that is similar in length to the note that explains why map::find( ) does not return an iterator?  (BTW, I like the logic there for returning a pointer, given limitations on std::optional).

fweik:
Added find method to set, analogous to map. Fixes #121.

### DIFF
--- a/immer/set.hpp
+++ b/immer/set.hpp
@@ -62,6 +62,11 @@ class set
 {
     using impl_t = detail::hamts::champ<T, Hash, Equal, MemoryPolicy, B>;
 
+    struct project_value_ptr
+    {
+        const T* operator()(const T& v) const noexcept { return &v; }
+    };
+
 public:
     using key_type        = T;
     using value_type      = T;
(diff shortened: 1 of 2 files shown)


arximboldi:
Cool, thanks a lot for your contribution! Merging as soon as it's green on Travis!

PR #153, opened 2020-07-29
Fix OSS-Fuzz issue 24466

arximboldi:
Introduced in 143f6ddcbf7e3c4f6660d34ab41dadcc375c8f7


arximboldi:
Ping @fedormatantsev do you object with this fix?

fedormatantsev:
Hey, @arximboldi , you're bringing back 2 warnings (msvc2019). `idx1` and `idx2` are `size_t`, then you try to convert it to `count_t`

arximboldi:
This code is however correct. `idx` has to be `size_t`, `count_t` is too small to hold the value, until we do `idx >> shift` below. (The warning may be solved with a cast down there, I don't care too much until we set up AppVeyor things will come back at different points).

Issue #160, opened 2020-09-03
Optional disabling of exceptions


chuim:
Our use of immer requires that we disable exceptions. As not supporting exceptions is a common situation in C++ codebases I'm filing this issue to propose integrating that change into immer.

Our modification follows the same pattern that [LibC++](https://github.com/llvm/llvm-project/tree/master/libcxx) uses:  if exceptions are enabled, use them, otherwise, if an out-of-range or an allocation fails, log a message and abort.

We added this snippet to `immer/immer/config.hpp`:
```C++
#if __has_feature(cxx_exceptions)
#define IMMER_THROW_BAD_ALLOC(msg) throw std::bad_alloc{msg}
#define IMMER_THROW_OUT_OF_RANGE(msg) throw std::out_of_range{msg}
#define IMMER_TRY try
#define IMMER_CATCH_ALL catch (...)
#define IMMER_RETHROW throw
#else
#define IMMER_NO_EXCEPTIONS
#define IMMER_THROW_BAD_ALLOC(msg) fprintf(stderr, "%s", msg); abort()
#define IMMER_THROW_OUT_OF_RANGE(msg) fprintf(stderr, "%s", msg); abort()
#define IMMER_TRY
#define IMMER_CATCH_ALL if (false)
#define IMMER_RETHROW
#endif
```
And then replaced all exception handling code with the appropriate macros.

Would it be OK to send a pull request with these modifications?

arximboldi:
That sounds good!

I would however change it for:
```cpp
#if __has_feature(cxx_exceptions)
#define IMMER_THROW(expr) throw expr
#define IMMER_TRY try
#define IMMER_CATCH(expr) catch (expr)
#define IMMER_RETHROW throw
#else
#define IMMER_NO_EXCEPTIONS
#define IMMER_THROW_EXPR(msg) \
    do {                     \ 
        assert(! #expr); \
        std::terminate(); \
    } while(false)
#define IMMER_TRY
#define IMMER_CATCH(expr) if (false)
#define IMMER_RETHROW
#endif
```

Any objections against these macros instead?

chuim:
Thanks for the reply. The only issue I see with using `assert` instead of a print function is that no error message would be printed on release builds (if `NDEBUG` is defined). But maybe that was your intention with this suggestion?

yesudeep:
[...]

PR #162, opened 2020-09-30
Make excpetions optional depending on build-time feature checks.

chuim:
Replaces all try/catch/throw keyword usage with a set of macros that are
appropriately defined in config.hpp based on the status of the
`cxx_exceptions` compiler feature.

Reported in issue #160.

### DIFF
--- a/immer/config.hpp
+++ b/immer/config.hpp
@@ -18,6 +18,28 @@
 #endif
 #endif
 
+#ifdef __has_feature
+#if !__has_feature(cxx_exceptions)
+#define IMMER_NO_EXCEPTIONS
+#endif
+#endif
+
+#ifdef IMMER_NO_EXCEPTIONS
+#define IMMER_TRY if (true)
+#define IMMER_CATCH(expr) else
+#define IMMER_THROW(expr) \
+    do {                  \
+        assert(! #expr);  \
+        std::terminate(); \
+    } while(false)
+#define IMMER_RETHROW
+#else
+#define IMMER_TRY try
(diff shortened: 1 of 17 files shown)


immer/config.hpp
```diff
@@ -18,6 +18,28 @@
 #endif
 #endif
 
+#ifdef __has_feature
+#if !__has_feature(cxx_exceptions)
```
chuim, line 22: I tried having these two checks in a single line to directly choose the version of the macros to use, but that caused build errors like this:
```
/usr/local/google/home/carlosk/lib/immer/immer/config.hpp:21:45: error: missing binary operator before token "("
   21 | #if !defined(__has_feature) || __has_feature(cxx_exceptions)
      |    
```
Splitting into two separate lines like this worked.

arximboldi (commented):
Overall LGTM. Have you run `clang-format` on it?

immer/detail/arrays/no_capacity.hpp
```diff
@@ -159,10 +159,13 @@ struct no_capacity
     no_capacity push_back(T value) const
     {
         auto p = node_t::copy_n(size + 1, ptr, size);
-        IMMER_TRY {
+        IMMER_TRY
+        {
```
arximboldi, line 163: The new line before the brace looks weird. Do you think there is any way to convince `clang-format` to move the brace back into the previous line?

arximboldi:
Cool, thanks! I think the last incarnation is acceptable. Merging!

PR #164, opened 2020-10-05
Added missing <immer/config.hpp> and <cstddef> includes

chuim:
The former needed for the new exception macros and the latter for the
usage of std::size_t.

Minor include ordering changes too.


arximboldi:
LGTM, thanks! :+1:

PR #179, opened 2021-05-01
Change <exception> includes to <stdexcept> when using standard exceptions

Yohsi:
Standard exceptions like `std::out_of_range` are defined in `<stdexcept>`, not in `<exception>`.
This fixes compilation issues i had with MSVC if I didn't include `<stdexcept>` before including some immer headers.
Thanks for this library btw !


arximboldi:
Excellent! Thank you!

Issue #183, opened 2021-07-23
Add key template parameter for map::find to allow look up with compatible objects


maierlars:
This allows lookup using a `std::string_view` while the keys are `std::string`s.

arximboldi:
Makes sense!

PR #184, opened 2021-07-23
Transparent Hash for Map and Sets

maierlars:
I opened this [issue](https://github.com/arximboldi/immer/issues/183) but then I though I could give it a try myself.

What I did was the following: whenever you refer to `K` as key type I replaced it with a template parameter. This allows you to put in any type that is supported by the hash object. I followed the c++20 convention for `std::unordered_map` such that those new functions only participate in overload resolution if `Hash::is_transparent` is a valid type.

This is just a proof of concept. There are more things one could consider:
- add template overloads to other member functions like `erase`.
- write more tests for other functions as well 🙈 

I added `immer/detail/rbts/rrbtree.hpp` because newer compilers no longer include `<limits>` as part of other stl headers.

### DIFF
--- a/immer/detail/rbts/rrbtree.hpp
+++ b/immer/detail/rbts/rrbtree.hpp
@@ -19,6 +19,7 @@
 #include <stdexcept>
 #include <memory>
 #include <numeric>
+#include <limits>
 
 namespace immer {
 namespace detail {
--- a/immer/map.hpp
+++ b/immer/map.hpp
@@ -114,7 +114,8 @@ class map
     {
         auto operator()(const value_t& v) { return Hash{}(v.first); }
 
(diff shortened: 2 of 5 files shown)


maierlars:
@arximboldi Please let me know if you are interested in having this feature added to immer.

arximboldi:
Hi @maierlars! I recently came back from an off the grid vacation and I'm slowly catching up with work and code stuff. I'll try take a look some time next week. Thank you a lot for this contribution, it's really appreciated!

arximboldi (approved):
Hi! Thank you very much for this contribution.  I very much aprreciate that you also included tests!

So sorry I didn't get to review it before.

Issue #120, opened 2020-02-03
State of diffing/patching redux


kuukunen:
Hi. Just wanted to check if there's any update for plans for diffing and/or patching support?
I read through the previous issues I found, #58 and #67, but the discussions died down a bit and you were wondering about use cases.

Basically we are currently evaluating using immutable data types and Redux-style architecture for creating an editor for our games.

So one use case is tens/hundreds of thousands of game objects in some combination of maps and arrays.

Another would be for example painting the shape/texture/blend values of a huge terrain, say, 16 km x 16 km. In the terrain case of course the actual data would be in just plain byte buffers, but for undo and optimization/streaming from disk it needs to be chunked in some manageable sizes. (Example of a tool: https://youtu.be/YPR1XFOUvp4?t=361 )

Part of the solution might have to have some sort of spatial data structure on top, but even still the chunks might be pretty big for efficient handling, otherwise the spatial bookkeeping becomes an issue and it would be a waste to write some sort of structural sharing system in addition to the existing one.

(Slightly contrived) example of a demanding operation: Artist is editing a forest that is part of a 16km x 16km world. They box select a thousand trees, then they move them a bit. Ideally the moving would be smooth and the trees would conform to the terrain.

We're actually just starting with that part, so we currently don't have specifics or any real world data, though.

Making an actual diff/patch functionality might of course be a bit gnarly, since you have to handle additions, deletions, changes, etc. There is some prior art in JS world, though:
https://github.com/intelie/immutable-js-diff
https://github.com/elierotenberg/remutable

I think for our use case the actual patching is not necessary, so one could do it with a diff iterator (possibly dereferencing, to a tuple, at least in a map case (Key1, Value1*, Value2*)). that could be used with a range based for.

I used pointers, but could also be (std::)optional. So if Value1 is nullptr/nullopt, you know Value2 was added, if vice versa, it was deleted, otherwise it was changed.

Or it could be its own special function, using a visitor, similar to for_each_chunk. And actually for performance reasons, especially for vector, it might even be necessary to have a for_each_diff_chunk that gives you all the chunks that *might* be different, based on the structural sharing, and leaves the element-wise diffing to the user.

It does become a bit complicated... Here's a quick sketch of usage...

```cpp
immer::vector<Object> state;
...
state.for_each_diff_chunk(oldState, [](size_t index, auto begin1, auto end1, auto begin2, auto end2) {
    auto it1 = begin1; // Or you could just use begin1 directly, but that's a bit hacky.
[...]

PR #199, opened 2022-01-26
Add diffing algorithm to maps

mvtec-richter:
This is the implementation of a diff algorithm for immer::map. It compares two immer::map instances and calls respective lambda functions when keys were added or removed or when the value of a key changed. When data is shared among the two maps, the algorithm will not need to compare shared subtrees of the map.
We plan to use this for an undo feature.

The diff algorithm has following signature:
̂`map::diff(map const& other, AddedFn&& added, ChangedFn&& changed, RemovedFn&& Removed)`
 where the lambdas have the signature:
```
void AddedFn(std::pair<Key, Value> const& added_entry)
void ChangedFn(std::pair<Key, Value> const& old_item, std::pair<Key, Value> const& new_data)
void RemovedFn(std::pair<Key, Value> const& removed_entry)
```
Usage:
```
immer::map<int, int> old_snapshot; // initialize somehow
immer::map<int, int> new_snapshot = old_snapshot.set(1,2); // change somehow
old_snapshot.diff(
  new_snapshot,
  [](auto const& added){ std::cout << "added key " << added.first << std::endl; },
  [](auto const& old, auto const& new){
    std::cout << "changed key " << old.first << " from " << old.second << " to " << new.second << std::endl; },
  [](auto const& removed){ std::cout << "removed key " << removed.first << std::endl; }
  );
```
This signature works for our usecase, however we are happy for further input and would like to help to upstream this feature.

This partially addresses: #120

### DIFF
--- a/immer/algorithm.hpp
+++ b/immer/algorithm.hpp
@@ -208,6 +208,74 @@ bool all_of(Iter first, Iter last, Pred p)
     });
 }
 
+template <class AddedFn, class RemovedFn, class ChangedFn>
+struct differ
+{
+    AddedFn added;
+    RemovedFn removed;
+    ChangedFn changed;
+};
+
+template <typename T>
+struct noop_changed
+{
+    void operator()(const T&, const T&) {}
+};
+
+template <class AddedFn, class RemovedFn, class ChangedFn>
+auto make_differ(AddedFn&& added, RemovedFn&& removed, ChangedFn&& changed)
+    -> differ<std::decay_t<AddedFn>,
+              std::decay_t<RemovedFn>,
+              std::decay_t<ChangedFn>>
+{
+    return {std::forward<AddedFn>(added),
+            std::forward<RemovedFn>(removed),
+            std::forward<ChangedFn>(changed)};
+}
+
+template <class AddedFn, class RemovedFn, class T>
+auto make_differ(AddedFn&& added, RemovedFn&& removed)
+    -> differ<std::decay_t<AddedFn>, std::decay_t<RemovedFn>, noop_changed<T>>
+{
+    return {std::forward<AddedFn>(added),
+            std::forward<RemovedFn>(removed),
+            noop_changed<T>{}};
+}
+
+template <typename T, typename Differ>
+void diff(const T& a, const T& b, Differ&& differ)
+{
+    a.impl().template diff<Differ, std::equal_to<T::value_type>>(
+        b.impl(), std::forward<Differ>(differ));
+}
+
+template <typename T, typename AddedFn, typename ChangedFn, typename RemovedFn>
+void diff_with(const T& a,
+               const T& b,
+               AddedFn&& added_fn,
+               RemovedFn&& removed_fn,
+               ChangedFn&& changed_fn)
+{
+    diff(a,
+         b,
+         std::move(make_differ(std::forward<AddedFn>(added_fn),
+                               std::forward<RemovedFn>(removed_fn),
+                               std::forward<ChangedFn>(changed_fn))));
+}
+
+template <typename T, typename AddedFn, typename RemovedFn>
+void diff_with(const T& a,
+               const T& b,
+               AddedFn&& added_fn,
+               RemovedFn&& removed_fn)
+{
+    diff(a,
+         b,
+         make_differ<AddedFn, RemovedFn, T::value_type>(
+             std::forward<AddedFn>(added_fn),
+             std::forward<RemovedFn>(removed_fn)));
+}
+
 /** @} */ // group: algorithm
 
 } // namespace immer
--- a/immer/detail/hamts/bits.hpp
+++ b/immer/detail/hamts/bits.hpp
@@ -125,6 +125,51 @@ inline count_t popcount(std::uint8_t x)
     return popcount(static_cast<std::uint32_t>(x));
 }
 
+template <typename bitmap_t>
+class set_bits_range
+{
+    bitmap_t bitmap;
+
+    class set_bits_iterator
+    {
+        bitmap_t bitmap;
+
+        inline static bitmap_t clearlsbit(bitmap_t bitmap)
+        {
+            return bitmap & (bitmap - 1);
+        }
+
+        inline static bitmap_t lsbit(bitmap_t bitmap)
+        {
+            return bitmap ^ clearlsbit(bitmap);
+        }
+
+    public:
+        set_bits_iterator(bitmap_t bitmap)
+            : bitmap(bitmap){};
+
+        set_bits_iterator operator++()
+        {
+            bitmap = clearlsbit(bitmap);
+            return *this;
+        }
+
+        bool operator!=(set_bits_iterator const& other) const
+        {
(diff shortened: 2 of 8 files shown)


arximboldi (changes requested):
Besides the questions above, I just saw one issue with the implementation:

Inside `champ`, you are accessing `.first` and `.second` on the values of the map. However, `champ` itself is agnostic of whether the tree represents a Map or a Set. You can see how `immer::detail::champ` is instantiated differently in `immer::set<>` and `immer::map<>` in order to achieve the two structures. I believe that you can achieve the desired genericity by using `Equal{}` instead of `a.first == b.first` and passing an additional function type to compare values (like we do for `champ::equals`).

Extra points if you also add tests for `immer::set` :smile: 

---

But then again, thank you so much for this brilliant PR!

mvtec-richter:
Thanks for the comments!

1. I agree a free function might be better. We had this in mind too. I'll implement this.
2. Grouping the function in an object is very sensible. I just wasn't sure about how to correctly do this with templates. I'll try your suggestion.
3. - 5. As we are not using vectors and lager, we haven't looked into this, but it sounds interesting.
I was aware of the issue with the std::pair in `immer::detail::champ` (sorry, I forgot to mention it in the PR). I will add another template parameter for comparing values to champ and see if I can get `immer::set` to work.

arximboldi:
Regarding the last point @mvtec-richter, you do not need to add another template parameter to `champ`, you can add it to the implementation of the diff function there. See `equals` as an example.

mvtec-richter:
I implemented the usage of a template variable for Equals and EqualsValue and also added sets.
Concerning changing the interface to a free function and using a struct for the function, I will have a look into it maybe tomorrow.
I noticed, that sets don't have a changed function, so these differ structs might be different for map/set/...

arximboldi:
@mvtec-richter thank you a lot for the latest changes!

I've noticed that you have added a `EqualValue` parameter to the `diff()` method of the public interface itself (`immer::map::diff`). I think we could for now not parametrize it and just use the default one; this makes it consistent with `immer::map::operator==`, which also just uses the default one. This means that there is no way to customize equality for values for now, but I am happy to live with that rigidity for simplicity.

Wrt. the `change` callback making set for sets, I wonder if it may make sense, to detect "changes" in values of two sets that compare equal when using `Equal`, but are different to `operator==`. What do you think?

mvtec-richter:
Of course I can remove the Equal template parameter from the interface. That will simplify the interface.

Wrt. the `change` callback, I don't think it makes sense to detect such "changes" in sets. If there are multiple layers of equal, you probably wanted a map in the first place.

However, if we have a free standing `diff` function, we could provide two variants, one with changed callback and one without. It might also be useful for maps to have a diff function without the changed callback. Even if it's not useful, I think its not completely wrong to allow a diff function with changed callback for sets.

I have another question concerning the use of std::forward for lambdas. I wonder, if our usage in champ::diff is correct. Should the callback be passed as & instead of &&, because they are forwarded multiple times in the function?

arximboldi:
Hi!

> Of course I can remove the Equal template parameter from the interface. That will simplify the interface.

Cool :+1: 

> Wrt. the change callback, I don't think it makes sense to detect such "changes" in sets. If there are multiple layers of equal, you probably wanted a map in the first place.

I agree :+1: 

> However, if we have a free standing diff function, we could provide two variants, one with changed callback and one without. It might also be useful for maps to have a diff function without the changed callback. Even if it's not useful, I think its not completely wrong to allow a diff function with changed callback for sets.

Yes, I agree that the interface should be homogeneous, so one can write generic code that uses `diff` regardless of whether the underlying data-structure is a a `map` or a `set` (and maybe in the future, even vector). I am happy with only having the 3-lambda version (or `differ` version) and having the `change` callback simply never do anything for `set`. Having the 2-lambda version additionally is also fine. I am happy with whatever you prefer in this case; with maybe a slight preference for having the 3-lambda or `differ` version only, since it's easier to add things than to remove things.

> I have another question concerning the use of std::forward for lambdas. I wonder, if our usage in champ::diff is correct. Should the callback be passed as & instead of &&, because they are forwarded multiple times in the function?

You are correct in that it's incorrect :) Taking the lambdas in as && is fine, but the `std::forward` call is what makes it incorrect. (Or more pedantically, it's not fully incorrect, it's only dangerous, because you are not really consuming the forwarded value by either forwading into the `operator()` nor forwading it into a by-value nor any r-value consuming function of any sort). SImply removing the `std::forward<>` calls in the tree traversal would fix the code. You don't need to remove it in the code that passes the lambdas down only once.

arximboldi:
I'm merging this and will work on polishing the details myself. Thank you very much for this work Florian!

khiner:
Hi @arximboldi ! I've been holding onto some pretty nasty architecture in my application, waiting and hoping for someone to extend the diff mechanism to `immer::set`, `immer::vector`, and `immer::flex_vector` 😅 

I'm hitting the point in my app where it's getting silly to work around this, so I've decided to block further development on having the ability to diff immer sets & vectors. Before jumping into implementation to contribute these diffs (which I've put off since it seems a bit intimidating...), I wonder if you have any thoughts/tips/suggestions, or on the outside chance you or someone else has already started to look into this?

It sounds like the current map diff implementation would be a good starting point?
> ... the interface should be homogeneous, so one can write generic code that uses diff regardless of whether the underlying data-structure is a a map or a set (and maybe in the future, even vector).

Thank you!

arximboldi:
Hi @khiner!

Thank you for looking into this... that would be an incredibly useful contribution. In fact... I do miss this feature all the time!

One of the reasons it hasn't been tackled though is that it's actually not easy work. It is definitely harder than HAMT diffing. In fact, I suspect that it would require some degree of original algorithmic work (potentially worth an academic paper though, in case that's an encouragement for you!).

As a starting point, I would recommend getting some familiarity with how Relaxed Radix Balanced Tries work. You can check the first parts of the [Immer paper](https://public.sinusoid.es/misc/immer/immer-icfp17.pdf). I also find [L'Orange's master's thesis](https://hypirion.com/thesis.pdf) one of the clearest descriptions of this data-structure.

Then we'd need to find some way to adapt some of the existing sequence diffing algorithms to leverage structural sharing, by operating somewhat recursively in the RRB-trie. You can look into [this Rust library](https://github.com/mitsuhiko/similar) for a list of existing algorithms (and their implementation). I have [this discussion](https://news.ycombinator.com/item?id=13983085) in my notes also as it contains interesting pointers in that field.

Let me know if there is some way I can further help!

khiner:
Hi @arximboldi , thanks for getting back on this and for the detailed notes and pointers! I see, it sounds like this is far from trivial, but in theory possible to achieve the same type of `O(|diff|)` complexity currently provided by the HAMT differ, in your estimation? This does sound like a rewarding project, but I may be too busy with other commitments to contribute this level of research... but I am motivated to have this, so maybe ;)

arximboldi:
Yes, my intuition says we we should be able to achieve that! Big big reward indeed! Game changer for lots of applications.

PR #201, opened 2022-02-09
Fix typo in documentation of diff

mvtec-richter:
Thanks for taking the time to polish the diff feature!


arximboldi:
Ooops, good catch!

PR #205, opened 2022-04-01
immer::array fix to prevent breakage with glm library

nghiaho12:
Tested with the following code

```cpp
#include <glm/glm.hpp>
#include "immer/array.hpp"

int main() {
    immer::array<glm::vec3> pos = {
        {0.f, 0.f, 0.f}
    };
}`
```

The error is
```
./immer/detail/arrays/with_capacity.hpp:131:22: error: invalid static_cast from type ‘const glm::vec<3, float, (glm::qualifier)0>*’ to type ‘immer::detail::arrays::with_capacity<glm::vec<3, float, (glm::qualifier)0>, immer::memory_policy<immer::free_list_heap_policy<immer::cpp_heap>, immer::refcount_policy, immer::spinlock_policy> >::size_t {aka long unsigned int}’
         auto count = static_cast<size_t>(distance(first, last));
```

glm provides a distance function which gets used by immer, somehow ...

### DIFF
--- a/immer/detail/arrays/no_capacity.hpp
+++ b/immer/detail/arrays/no_capacity.hpp
@@ -108,7 +108,7 @@ struct no_capacity
                                bool> = true>
     static no_capacity from_range(Iter first, Sent last)
     {
-        auto count = static_cast<size_t>(distance(first, last));
+        auto count = static_cast<size_t>(std::distance(first, last));
         if (count == 0)
             return empty();
         else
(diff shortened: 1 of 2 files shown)


arximboldi:
It gets used because of ADL, to allow for iterators to customize it.

I feel like GLM is in the wrong here. Still my implementation looks a bit odd: there should be a `using std::distance` just before the use of it. I don't think that adding that would solve your issues though...

PR #206, opened 2022-04-21
Add Swift Package

wtholliday:
Adds a Swift Package so immer can be used easily in Xcode projects. SPM doesn't know about header-only libraries, so it required adding a source file (spm.cpp). We can move this file wherever you want.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,29 @@
+// swift-tools-version:5.5
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "immer",
+    products: [
+        // Products define the executables and libraries a package produces, and make them visible to other packages.
+        .library(
+            name: "immer",
+            targets: ["immer"]),
+    ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        // .package(url: /* package url */, from: "1.0.0"),
+    ],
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -0,0 +1,29 @@
...
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages this package depends on.
+        .target(
+            name: "immer",
+            dependencies: [],
+            path: ".",
+            sources: ["spm.cpp"],
+            publicHeadersPath: ".")
+    ],
+    cxxLanguageStandard: .cxx20
```
arximboldi, line 28: Immer is actually C++14 compatible, would it be possible to lower this?

spm.cpp
```diff
@@ -0,0 +1,4 @@
+
+#include <immer/vector.hpp>
+
+// Just so we can compile with SPM.
```
arximboldi, line 4: So there is no way to make a purely header only library for `spm`?

arximboldi (commented):
I've made a couple of comments. I think it could actually be useful to officially support SPM, and as a header only library with no dependencies, it should be almost trivial to keep it up to date. See the questions inline before we can merge though.

Package.swift
wtholliday: Yep!

Issue #198, opened 2022-01-26
Add maps with keys extracted from the value


arximboldi:
Solves the issue of:
```c++
struct something {
  id_t id;
   ...
};

using something_map = immer::map<id_t, something>;
```

arximboldi:
With this new type one could do something like:
```c++
struct something {
  std::string id;
   ...
};

auto x = immer::object_map<something>{}
   .insert(something{"foo", ...})
   .insert(something{"bar", ...}));

static_assert(std::is_same_v<decltype(x)::key_type, std::string>);
assert(x.find("foo") != nullptr);
assert(x.at("foo") == something{"foo", ...});
assert(x.count("foo") == 1);
```
We could say that by default it access the `id` or `key` attribute, but an arbitrary key projection could be used by using customization points:
```c++
struct something {
  int name;
  ...
};

struct something_key {
   int operator()(const something_key& x) { return x.name; }
};

using something_map = immer::object_map<something, something_key>{};
```
Not sure if `object_map` is the best name for this construction. Happy to hear other ideas!

deadb0d4:
Cool, let me try it. It indeed seems like a good first issue here :)

deadb0d4:
I skimmed through the `{map,set}.hpp`, I think about the following (don't mind my code style, I will adjust it when I am working on the PR):

```c++

struct DefaultIndexer {
  template <class T>
  auto operator() (const T& t) {
    return t.id;
  }
};

// it will be a bit tedious to define concept for Indexer w/o c++20
template <typename T, typename Indexer = DefaultIndexer>
class single_hash_index {
 private:
  using key_type = decltype(Indexer()(T()));

 public:
  // semantics of the immer::map

 private:
  immer::set<T> values;
};
```
The idea is that user can specialize `DefaultIndexer::operator()` as well as providing its own `Indexer`. @arximboldi do you see any issues with this? I would like to have someone's approve before working on PR...

arximboldi:
Yes, the idea is to provide a new type with an interface very similar to `immer::map`, and also based on `detail::champ` as implementation. You can achieve the right interface by customizing `project_value` and the other callbacks that are used with `champ`. It should be not too difficult if you are already familiar with templates and the Immer interface.

arximboldi:
Btw @deadb0d4, are you on Twitter? I'm gonna make an announcement about the new type, and I would be very happy to credit you with a direct mention!

deadb0d4:
[...]

PR #210, opened 2022-05-26
enhance: add hash map with keys extracted from values

deadb0d4:
This PR attempts to add new class `single_hash_index` with required functionality using `detail::champ` as implementation.

Issue: #198


deadb0d4:
**TODO:**

- Tests
- Docs

deadb0d4:
@arximboldi Could you please take the first look before my tests just in case something is off already?

immer/single_hash_index.hpp
```diff
@@ -0,0 +1,318 @@
...
+          detail::hamts::bits_t B>
+class single_hash_index_transient;
+
+template <class T>
+struct default_key_extractor
+{
+    using key_type = decltype(T().id);
+
+    key_type operator()(const T& t) const { return t.id; }
+};
```
arximboldi, line 25: I would suggest that the `default_key_extractor` delegate in a free function. That way the free function can be used as customization point in a way simpler than providing a custom type. This also allows us to provide multiple default syntaxes supported. This could look like this:
```c++
template <typename T>
auto get_key(const T& x) -> decltype(x.id) { return x.id; }

struct key_fn
{
    template <typename T>
    decltype(auto) operator()(const T& x) const { return get_key(x); }
};
```

Note also how this simplifies the definition of the extractor, not needing to make the whole type a template. In order to define `key_type` in the `table` itself, you can do:
```c++
using K = std::decay_t<decltype(KeyFn{}(std::declval<T>()))>;
```

immer/single_hash_index.hpp
```diff
@@ -0,0 +1,318 @@
...
+
+template <typename T,
+          typename KeyExtractor = default_key_extractor<T>,
+          typename Hash         = std::hash<typename KeyExtractor::key_type>,
+          typename Equal = std::equal_to<typename KeyExtractor::key_type>,
+          typename MemoryPolicy   = default_memory_policy,
+          detail::hamts::bits_t B = default_bits>
+class single_hash_index
```
arximboldi, line 33: I have been giving it some thought, and I think that `immer::table<>` would be a good name for this type. What do you think?

I don't like `single_hash_index`, it is very long and a bit confusing about what it means.

arximboldi (changes requested):
This looks pretty nice, thanks a lot!

There are some changes requested, please see the comments inline.  Also, before this is merged, both tests and documentation will be required.

Thank you very much!!!

deadb0d4:
@arximboldi I've made a rough version of tests & docs. Mainly, I've copy-pasted stuff from `map/map_transient`.

arximboldi (changes requested):
This is looking fantastic! :heart: 

Just a couple of missing things:

- Please add the class [here](https://github.com/arximboldi/immer/blob/master/doc/containers.rst) in order for it to appear in the documentation page.

- One important thing about the `update()` function: the `fn` *must* not change the `id`! That is probably worth mentioning in the documentation for the function.

deadb0d4:
There is a weird error in the CI, is it known?
```
0s
Run shimataro/ssh-key-action@v2
  with:
    name: id_rsa
    if_key_exists: fail
  env:
    NIX_PATH: nixpkgs=channel:nixos-unstable
Error: Input required and not supplied: key
```

arximboldi:
That CI error can be ignored. It is because CI secrets are hidden for external contributors, so it is not applying the SSH key.

PR #217, opened 2022-06-09
HAMT transients: move mutable inner nodes when posible during insertion

arximboldi:
In previous PR's we optimized "mutating" insertions by changing nodes in place when possible. However, whenever the change in-place was not possible because the new node needed is of a different size, the operation was performed exactly like in the immutable case.

This PR implements the realization that, when the node needs to be to be reallocated, we can still do things in a significantly faster way. In that case, we can "move" things out of the old node. When using reference counting, this means skipping touching the reference counts at all for inner nodes. For value buffers, we use std::move() potentially saving further allocations or reference count touches.

In some local testing, this brings the performance of mutating operations to 1.5X to 2.5X the performance of std::unordered_map, depending on various scenarios. This is pretty good!

This implements this optimization for insert(), set() and update(). For erase() it will come in a follow up PR.

FYI @omer-s @harryhk

### DIFF
--- a/immer/config.hpp
+++ b/immer/config.hpp
@@ -8,6 +8,10 @@
 
 #pragma once
 
+#if (__cplusplus >= 201703L) || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
+#define IMMER_HAS_CPP17 1
+#endif
+
 #if defined(__has_cpp_attribute)
 #if __has_cpp_attribute(nodiscard)
 #define IMMER_NODISCARD [[nodiscard]]
--- a/immer/detail/arrays/node.hpp
+++ b/immer/detail/arrays/node.hpp
@@ -61,7 +61,7 @@ struct node
 
     static void delete_n(node_t* p, size_t sz, size_t cap)
     {
-        destroy_n(p->data(), sz);
+        detail::destroy_n(p->data(), sz);
         heap::deallocate(sizeof_n(cap), p);
     }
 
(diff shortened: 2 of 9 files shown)


arximboldi:
Some quick links to benchmarks, run by the Github Actions.

They show producing sets of 1000 elements with various types (`immer::box<std::string>`, `std::string` both short and long, and `unsigned int`).

As mentioned yesterday, it seems that we now are at around 1.5X and 2.5X `std::unordered_map`, which is pretty neat :smile: 

**Clang**
- https://public.sinusoid.es/misc/immer/reports/report_9caf111_fv-az371-66_clang++_N:1000_s20/benchmark-set-string-box-insert.html
- https://public.sinusoid.es/misc/immer/reports/report_9caf111_fv-az371-66_clang++_N:1000_s20/benchmark-set-string-long-insert.html
- https://public.sinusoid.es/misc/immer/reports/report_9caf111_fv-az371-66_clang++_N:1000_s20/benchmark-set-string-short-insert.html
- https://public.sinusoid.es/misc/immer/reports/report_9caf111_fv-az371-66_clang++_N:1000_s20/benchmark-set-unsigned-insert.html

**GCC**
- https://public.sinusoid.es/misc/immer/reports/report_9caf111_fv-az36-48_g++_N:1000_s20/benchmark-set-string-box-insert.html
- https://public.sinusoid.es/misc/immer/reports/report_9caf111_fv-az36-48_g++_N:1000_s20/benchmark-set-string-long-insert.html
- https://public.sinusoid.es/misc/immer/reports/report_9caf111_fv-az36-48_g++_N:1000_s20/benchmark-set-string-short-insert.html
- https://public.sinusoid.es/misc/immer/reports/report_9caf111_fv-az36-48_g++_N:1000_s20/benchmark-set-unsigned-insert.html

PR #218, opened 2022-06-11
Fix -fno-exceptions build (regression from 35147686)

cyrilRomain:


arximboldi:
Ooops! Thank you! We should add that to CI so it catches it...

PR #221, opened 2022-06-28
Add optional method to print statistics about CHAMP density and utilization

arximboldi:
With this, when definining `IMMER_DEBUG_STATS=1` one can do:
```c++
v.impl().get_debug_stats().print();
```
to get statistics about utilization in HAMT-based data-structures. This can be used to inform whether it is a good idea to preallocate nodes fully.

This branch also contains another fix for a memory leak found by oss-fuzz, the last one I think!


arximboldi:
FYI @omer-s @harryhk

PR #233, opened 2022-10-13
make clang ignore gcc pragma warning for champ

asa:
Clang throws a warning about this unknown gcc pragma. This is bad when we have warnings -> errors on.


arximboldi (approved):
The warnings game! Thanks Asa!

Issue #244, opened 2022-12-29
`noexcept` move constructors for `flex_vector`


maierlars:
I looked into making the move constructor (and move-assignment) for `flex_vector` `noexcept`. This is something people (at least myself) want to ensure exception safety. Generally you want you types to be nothrow move constructible. 

It turns out to be not that trivial: it allocates nodes. This is similar to what the MSVC implementation of `std::unordered_map` does and it caused me a lot of pain in the past. At least the `swap` operation is de facto noexcept, but not marked as such.

I propose two solutions:
1. do not allocate anything unless necessary. This would introduce some special cases, where `nullptr` to head and tail node would indicate an empty flex_vector.
2. This is similar to libstdc++ does for `std::unordered_map`: store the end nodes within the object itself. This would require to replace the end-nodes with the new end-nodes during a move operation.

I prefer approach 1. Any ideas or opinions? Maybe there is a good reason to not have flex_vectors move constructor and assignment noexcept.

arximboldi:
Hmmm... it does allocate memory but only the first time it is ever invoked, to allocate the initial "empty" node. We could maybe add some statics to pre-initialize the empty node... but that could cause trouble when using the data-structures in statics itself. Maybe the option 1) is the best? Not sure about how 2) would look like (what's an end node? you mean the root node?). I agree having a `noxexcept` move can be rather crucial.

maierlars:
I see, it's essentially this [piece](https://github.com/arximboldi/immer/blob/master/immer/detail/rbts/rrbtree.hpp#L50-L54) of code, that does an allocation.

Since the nodes are already shared between multiple flex-vectors, instead of keeping a pointer to an allocation in a static variable, one could instead keep the whole node as a static variable and return a pointer to it. 🤔 I might try that later today.

maierlars:
https://github.com/arximboldi/immer/pull/246

maierlars:
```
    /*!
     * Default constructor.  It creates a flex_vector of `size() == 0`.
     * It does not allocate memory and its complexity is @f$ O(1) @f$.
     */
    flex_vector() = default;
```
I just noticed that the comment is _slightly_ incorrect.

arximboldi:
I love the approach with a static allocation for the node. It's the cleanest approach. More comments in the PR, but thanks a lot for noticing and taking care of this!

arximboldi:
Renamed and left it open to keep track of this for other data-structures. Thanks!

[...]

PR #246, opened 2022-12-30
`flex_vector` nothrow move constructible/assignable

maierlars:
This is a result of https://github.com/arximboldi/immer/issues/244.

This is just an idea how one could make the move constructor and assignment operator `noexcept`.
I used `aligned_storage` to statically reserve some memory for the nodes.

However, while I type this description I noticed a problem: I'm not familiar with the mechanism the frees memory. There is nothing that prevents the code from attempting to free the statically allocated buffers.
A possible solution would be to keep the ref count always above 0.

### DIFF
--- a/immer/detail/rbts/node.hpp
+++ b/immer/detail/rbts/node.hpp
@@ -330,7 +330,7 @@ struct node
     static node_t* make_leaf_n(count_t n)
     {
         assert(n <= branches<BL>);
-        auto m  =heap::allocate(sizeof_leaf_n(n));
+        auto m = heap::allocate(sizeof_leaf_n(n));
         return make_leaf_n_into(m, sizeof_leaf_n(n), n);
     }
 
--- a/immer/detail/rbts/node.hpp
+++ b/immer/detail/rbts/node.hpp
@@ -199,16 +199,22 @@ struct node
     }
     static ownee_t& ownee(node_t* x) { return get<ownee_t>(x->impl); }
 
-    static node_t* make_inner_n(count_t n)
+    static node_t* make_inner_n_into(void* buffer, std::size_t size, count_t n)
     {
         assert(n <= branches<B>);
-        auto m                       = heap::allocate(sizeof_inner_n(n));
-        auto p                       = new (m) node_t;
+        assert(size >= sizeof_inner_n(n));
+        auto p                       = new (buffer) node_t;
         p->impl.d.data.inner.relaxed = nullptr;
 #if IMMER_TAGGED_NODE
         p->impl.d.kind = node_t::kind_t::inner;
 #endif
         return p;
+    }
+    static node_t* make_inner_n(count_t n)
+    {
+        assert(n <= branches<B>);
+        auto m = heap::allocate(sizeof_inner_n(n));
+        return make_inner_n_into(m, sizeof_inner_n(n), n);
     }
 
     static node_t* make_inner_e(edit_t e)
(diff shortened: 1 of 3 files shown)


immer/detail/rbts/rrbtree.hpp
```diff
@@ -49,13 +49,18 @@ struct rrbtree
...
+            node_t::make_inner_n_into(&storage, size, 0u);
         return empty_->inc();
     }
 
     static node_t* empty_tail()
     {
-        static const auto empty_ = node_t::make_leaf_n(0u);
+        constexpr auto size = node_t::sizeof_leaf_n(0);
+        static std::aligned_storage_t<size, alignof(std::max_align_t)> storage;
+        static const auto empty_ = node_t::make_leaf_n_into(&storage, size, 0u);
```
arximboldi, line 63: I think I would prefer this:
```c++
static const auto empty_ = [] {
  static std::aligned_storage_t<size, alignof(std::max_align_t)> storage;
  return node_t::make_leaf_n_into(&storage, size, 0u);
}();
```
This hides `storage` inside the initialization of `empty_` which I think could have some advantages (from an understandability POV, but it also makes it clear to the compiler that it only needs to access it once, during the protected initialization of `empty_` itself.)

arximboldi: The same self-executing lambda pattern can be used for the root above.

immer/detail/rbts/node.hpp
```diff
@@ -310,16 +316,24 @@ struct node
...
 #if IMMER_TAGGED_NODE
         p->impl.d.kind = node_t::kind_t::leaf;
 #endif
         return p;
     }
 
+    static node_t* make_leaf_n(count_t n)
+    {
+        assert(n <= branches<BL>);
+        auto m  =heap::allocate(sizeof_leaf_n(n));
```
arximboldi, line 333: I think the code has not be run through `clang-format`, can you do that?

arximboldi (changes requested):
There is a small comment inline, but in general, I really love this approach, I think this is the cleanest solution. 

Re. avoiding "deallocation", ensuring the refcount is always above 0 is the right approach. It seems you are already doing that, so that looks good.

Also, further improvements that you could optionally include in this PR, or if you want, tackle in later further PRs.
- It looks like copy-construction and assignment can also be made `noexcept`.
- I'd love to see the same pattern applied to all other data-structures, I believe `array`, `box`, `vector`, `map`, `set` and `table` all could benefit from this technique.

immer/detail/rbts/rrbtree.hpp
maierlars: Very good!

immer/detail/rbts/node.hpp
maierlars: Done

maierlars:
How can we proceed with this PR? :)

arximboldi:
I am happy to merge it. Just wasn't sure what your opinion is about the other improvements I suggested. I guess they could be tackled in a separate PR, would you like to do them? Should I just merge this branch as-is for now?

maierlars:
If you are happy with this PR, please merge it. For my work it is not urgent to have the other data-structures noexcept move-constr, however, I totally agree, that this should be done at some point. Maybe I find time to look into this over the weekend, but if someone else feels like doing it, please go ahead.

arximboldi:
Sounds good to me Lars, merged!

Also, quite interesting the use-case for Arango DB. Curious about how you are using the library and what benefits it is providing. Happy to follow up per e-mail some time if you'd like to keep that a bit more private: juanpe@sinusoid.es

maierlars:
Sure, there a multiple use cases for immer at ArangoDB, however, the flex vector is used primarily for an in-memory representation of a sequence of operations (which is also dumped to disk). For each shard we keep the history of the recent operations on documents in that shard and replicated the operations from the leader to its followers. While new log entries are appended, old entries are eventually compacted (removed) and we take snapshots of ranges for iteration or network messages regularly.

At the time we initially designed the prototype it just felt natural to use an immutable data structure for that. And until today we see no performance loss. After all, we are writing a database and want to replicate data, so the whole business is clearly dominated by disk and network io times. In fact, I personally have less mental load, I can just create a copy: it's cheap, it's not modified, its' noexcept, and no thread is blocked by that.

PR #247, opened 2023-01-05
Fix wrong order in operator<(T2, box<T>)

urbanwe:


arximboldi (approved):
Good catch. Thanks a lot!

Issue #57, opened 2018-09-10
CMake enhancements


apmccartney:
Hello @arximboldi 

I would like to embellish the CMake support provided by the library. Namely,

+ support Catch-style subproject composition
+ provide an installation target and CMake configuration exports, to simplify downstream consumption 
+ provide integration with CTest, to provide enhanced testing facilities and result collection
+ provide integration with CPack, to support the generation of installation packages, e.g. 
  + `deb` and `rpm` archives for Linux 
  + `dmg` images for macOS
  + NSIS installation wizards for Windows

I've contributed this sort of support to several other projects. See [frozen](https://github.com/serge-sans-paille/frozen) and [stlab](https://github.com/stlab/libraries) for examples of what this ends up looking like.

arximboldi:
That sounds good. I am happy to take in contributions in that direction.

I have some concerns though:
- There is already an installation target. In which way is it not suitable?
- There is already ctest enabled tests. In which way are they not suitable?
- What is exactly does "support Catch-style subproject composition" mean?
- Awesome to have cpack support!

apmccartney:
> There is already an installation target. In which way is it not suitable?

'Not suitable' is too strong, but I think there is room for improvement.

The configuration file generated by CMake does not express external library dependencies, e.g. boost and boehm-gc. The former only seems necessary for the experimental `dvektor` and `testing`, but the latter might be relevant to consuming projects. 

Daniel Pfiefer's touched on this shortcoming and suggests a workaround in his [2017 C++Now CMake talk](https://www.youtube.com/watch?v=bsXLMQ6WgIk). In short, the CMake generated configuration file can be included in a small library provided config file that establishes those dependencies via
`find_dependency`.

Moreover, the configuration file generated by CMake does not express library version information or version compatibility. A second file, `<project name>Config-Version.cmake` is required to establish that sort of information. There is an intrinsic cmake module that provides functions to generate such a file, which can be installed with the cmake-generated config file.

> There is already ctest enabled tests. In which way are they not suitable?

Immer provides tests using `add_test` which can be consumed by CTest. You even respect composing projects by opting into testing via the `make check` target. This is not an uncommon convention in general and I've encountered it in a number of autotools projects. 

What's lost is a bit of idiomatic usage and CDash integration. 

[...]

PR #250, opened 2023-01-12
Add export CMake Config version file

LilyWangLL:
There is an vcpkg issue: https://github.com/microsoft/vcpkg/issues/28864, some users need use `find_package(immer VERSION 0.8.0)` feature. 

But this port doesn't have this feature. So I submit this PR to add export CMake Config version file, this will fix the use of find special version of `immer` like `find_package(immer VERSION 0.8.0)`, the parameter `SameMajorVersion` applies to the version with major 0.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,14 @@ install(TARGETS immer EXPORT ImmerConfig)
 install(EXPORT ImmerConfig DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/Immer")
 install(DIRECTORY immer DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
 
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -99,6 +99,14 @@ install(TARGETS immer EXPORT ImmerConfig)
 install(EXPORT ImmerConfig DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/Immer")
 install(DIRECTORY immer DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
 
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(
+  ${CMAKE_CURRENT_BINARY_DIR}/ImmerConfigVersion.cmake
```
dg0yt, line 104: Missing quotes.

dg0yt (changes requested):

arximboldi (approved):
LGTM!

Will merge after CI runs.

tocic:
This addresses one of the problems mentioned in #57.

Issue #137, opened 2020-06-28
Possibly address unused parameter/variable warnings


gorakhargosh:
Hi! 

Thank you for creating this wonderful library. I have a small request.
Could you please turn on -Wall -Wpedantic -Werror -Wunused and 
try to minimize the unused variables/parameters? We see a flurry of 
warnings when we build code that uses immer. 

```
INFO: From Compiling codelab/cc/containers/immutable/vector_test.cc:
In file included from codelab/cc/containers/immutable/vector_test.cc:1:
In file included from bazel-out/darwin-fastbuild/bin/external/com_github_arximboldi_immer/_virtual_includes/immer/immer/vector.hpp:11:
In file included from bazel-out/darwin-fastbuild/bin/external/com_github_arximboldi_immer/_virtual_includes/immer/immer/detail/rbts/rbtree.hpp:11:
In file included from bazel-out/darwin-fastbuild/bin/external/com_github_arximboldi_immer/_virtual_includes/immer/immer/detail/rbts/node.hpp:13:
bazel-out/darwin-fastbuild/bin/external/com_github_arximboldi_immer/_virtual_includes/immer/immer/detail/util.hpp:132:20: warning: unused parameter 'f' [-Wunused-parameter]
auto static_if(F&& f) -> std::enable_if_t<!b>
                   ^
bazel-out/darwin-fastbuild/bin/external/com_github_arximboldi_immer/_virtual_includes/immer/immer/detail/util.hpp:136:30: warning: unused parameter 'f2' [-Wunused-parameter]
auto static_if(F1&& f1, F2&& f2) -> std::enable_if_t<b, R>
                             ^
bazel-out/darwin-fastbuild/bin/external/com_github_arximboldi_immer/_virtual_includes/immer/immer/detail/util.hpp:141:21: warning: unused parameter 'f1' [-Wunused-parameter]
auto static_if(F1&& f1, F2&& f2) -> std::enable_if_t<!b, R>
                    ^
In file included from codelab/cc/containers/immutable/vector_test.cc:1:
In file included from bazel-out/darwin-fastbuild/bin/external/com_github_arximboldi_immer/_virtual_includes/immer/immer/vector.hpp:11:
In file included from bazel-out/darwin-fastbuild/bin/external/com_github_arximboldi_immer/_virtual_includes/immer/immer/detail/rbts/rbtree.hpp:11:
[...]

PR #252, opened 2023-03-03
fix some unused parameters

e-dant:
Addresses part of https://github.com/arximboldi/immer/issues/137

PS Thanks for your work

### DIFF
--- a/immer/detail/util.hpp
+++ b/immer/detail/util.hpp
@@ -232,16 +232,16 @@ auto static_if(F&& f) -> std::enable_if_t<b>
     std::forward<F>(f)(empty_t{});
 }
 template <bool b, typename F>
-auto static_if(F&& f) -> std::enable_if_t<!b>
+auto static_if() -> std::enable_if_t<!b>
 {}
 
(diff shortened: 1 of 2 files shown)


immer/detail/util.hpp
```diff
@@ -232,16 +232,16 @@ auto static_if(F&& f) -> std::enable_if_t<b>
     std::forward<F>(f)(empty_t{});
 }
 template <bool b, typename F>
-auto static_if(F&& f) -> std::enable_if_t<!b>
```
e-dant, line 235: Maybe these should just be annotated as maybe unused instead of removed

arximboldi: Sorry for the very late review! This changes the signature of the function so I don't think this work. Maybe you can do it like this:
```c++
template <bool b, typename F>
auto static_if(F&&) -> std::enable_if_t<!b>
```
??

arximboldi:
Sorry for the late review of the code. I'd be happy to merge it once the issues in the comment is addressed and the conflicts resolved.

Issue #97, opened 2019-08-31
GDB Pretty printing support


dwightguth:
I work for a startup that is developing an interpreter for a functional programming language, and we recently decided to start using this library as the implementation of our map, list, and set types in our language. At roughly the same time, I have been working on improving the support for our programming language, which compiles down to native code, in GDB.

One thing that was particularly useful was to be able to use gdb's `print` command to print the elements of a map, list, or set. I originally tried to make this work by invoking functions inside the inferior process, but it turns out gdb segfaults when you invoke code from within a custom pretty printer. As a result, it proved necessary to reimplement the logic of rrbtree_iterator and champ_iterator in python using GDB's python API.

The code I wrote works great, although it is not fully production ready for inclusion back into the library in the sense that it is still specialized in several places to the specific template parameters we use for our data types in our programming language, but I wanted to bring your attention to the code since it could potentially be used to drastically improve the experience of debugging other programs that make use of these collections.

[Here is the source code.](https://gist.github.com/dwightguth/283afe96b60b3793f3c02036701457f8)

arximboldi:
That is super cool! If you flesh it out into sth that you think could live in the main repo, feel free to make a PR. There is an `extra` and `tools` folders where something like this could live.

arximboldi:
Also, quite curious about this functional language. I hope Immer helps!

dwightguth:
Unfortunately it's unlikely that I'll have a chance to flesh this out the rest of the way anytime soon because the main thing that is left is to make this code properly parametric in the template parameters, and that's not something that is required for the use case that I'm dealing with because the template parameters are always the same in our code. I would have to spend some time on it as a hobby, and I have other hobby projects at the moment that would provide more utility to me.

I'm happy to answer questions about this code if it would help you finish the work yourself though, if you're interested in trying to do it.

The functional language is called [K](https://github.com/kframework/k) by the way and the code that uses immer is [here](https://github.com/kframework/llvm-backend/tree/master/runtime/collections) if you want to take a look. The library has turned out to be a great choice and it is significantly faster than our old implementation, which was based on a similar rust library. We would significantly benefit from map and set transients, since we use garbage collection and have to do a fair number of batch changes, but it's still quite good performance. And I really like how simple and clear the immer implementation is.

arximboldi:
I understand. I will leave the issue open, so your implementation is easier to find for people that are looking for one.

[...]

PR #255, opened 2023-04-17
Add GDB pretty-printing support

dhly-etc:
As discussed in #97, I'm submitting this PR, which adds GDB pretty-printing support. The code is sourced from @dwightguth's original [gist](https://gist.github.com/dwightguth/283afe96b60b3793f3c02036701457f8), and lightly modified. Dwight very kindly dual-licensed his gist under BSL/BSD to minimize any legal issues with the integration of this code.

I'd welcome any feedback, and would be happy to try and address any concerns.


dwightguth (commented):
I would look into the gdb auto load process for python scripts. I don't believe that this example, by itself, will be sufficient for people who use the library to get this functionality automatically. There are several different ways that python scripts can be loaded into gdb automatically, and you probably want to think about which one is best for your use case.

If I'm missing something and you've already tested this and it works without needing to be manually loaded, feel free to ignore my comment. I'll look at the code in more detail tomorrow.

tools/gdb_pretty_printers/printers.py
```diff
@@ -0,0 +1,432 @@
...
+
+    basename = match.group(1)
+    if basename == "immer::map":
+        return MapPrinter(val)
+    elif basename == "immer::set":
+        return SetPrinter(val)
+    elif basename == "immer::vector":
+        return ListPrinter(val, "immer::vector")
+    elif basename == "immer::flex_vector":
+        return ListPrinter(val, "immer::flex_vector")
```
arximboldi, line 431: Likewise, maybe `immer::array` with `ListPrinter`

arximboldi (commented):
I like this, thank you!

I haven't tried it yet, can you provide some instructions on how to use this? I am not familiar with gdb pretty printers.

Also @dwightguth what is that auto-load process?

dwightguth:
https://sourceware.org/gdb/onlinedocs/gdb/Python-Auto_002dloading.html#Python-Auto_002dloading

https://sourceware.org/gdb/onlinedocs/gdb/Auto_002dloading-extensions.html#Auto_002dloading-extensions

These documents and the links found in them describe two different ways to make gdb auto load a python script. For security reasons you're still probably going to have to tell gdb that you trust the path containing the script, but once a user does that, they will automatically load the pretty printers contained in this script. Otherwise you would have to manually load the python code each time you start gdb, or else add code to load it to the scripts that run when gdb starts.

tools/gdb_pretty_printers/printers.py
dwightguth: The printers in question are dependent on the runtime layout of the type in question, because they essentially have to reimplement the behavior of an iterator in python over the internal structure of the type. I believe array uses a different underlying representation, so a separate iterator would have to be added.

dhly-etc:
>If I'm missing something and you've already tested this and it works without needing to be manually loaded, feel free to ignore my comment. I'll look at the code in more detail tomorrow.

Nope, you're not missing anything. I've been testing by manually loading. In my case, I added a `source <path/to/autoload.py>` line to my `.gdbinit` file. You can also just type that directly into the gdb prompt to load the printers, if you don't have a `.gdbinit` file.

I'm looking into the autoloading options now. From a cursory read, the `.debug_gdb_scripts` method seems most promising given that this is a header-only library, but I haven't quite understood the details of this method yet.

tools/gdb_pretty_printers/printers.py
```diff
@@ -0,0 +1,432 @@
...
+    if not typename:
+        return None
+    match = compiled_rx.match(typename)
+    if not match:
+        return None
+
+    basename = match.group(1)
+    if basename == "immer::map":
+        return MapPrinter(val)
+    elif basename == "immer::set":
```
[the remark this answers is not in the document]
dhly-etc: Yep, this was pretty straightforward, so I've added it.

tools/gdb_pretty_printers/printers.py
dhly-etc: It does require a separate iterator, but it's pretty straightforward. I've added an array printer as well.

dhly-etc:
After looking more into `.debug_gdb_scripts`, I'm somewhat hesitant to try and tackle this form of auto-loading myself. It looks like it can be a bit tricky to get it right, especially across compilers.

I'm wondering if maybe it would be sufficient to add a `.gdbinit` file to the repo, so that anyone developing the library will get automatic loading, and anyone using the library in another project will at least have a hint as to how to use the printers. What do you think?

arximboldi:
That approach sounds good to me @dhly-etc !

dhly-etc:
@arximboldi Okay, in theory now if you open a `gdb` process from your project directory, you should automatically get the pretty printers. Let me know if this works for you!

Also let me know if you find any issues with the printers themselves. You should get legible output for the container contents. For instance, given a `immer::array<std::pair<std::string, int>> v3`, you should get output like
```
(gdb) print v3
$1 = immer::array with 3 elements = {{
    first = "a",
    second = 1
  }, {
    first = "b",
    second = 2
  }, {
    first = "c",
    second = 3
  }}
```

dwightguth:
I mean, telling people how to set it up themselves is a workable solution if you're concerned about the platform compatibility of the auto loading process. It can be somewhat finicky so I won't necessarily argue with your approach since it sounds like it's relatively simple to add the script to your gdbinit...

arximboldi:
This is very cool, thank you both for figuring this out.

CI is failing but for other reasons... let's go! :+1:

PR #256, opened 2023-04-28
Make pretty-printer loading more robust

dhly-etc:
A while after making the initial commit in #255, it came to my attention that `__file__` isn't actually a terribly reliable way to find the source filename. This change should make the loading process more robust, and work in more contexts.


dhly-etc:
@arximboldi Sorry to pester you, but hopefully this is a quick one to review?

arximboldi:
Sorry for the delays! Yes LGTM!

Issue #240, opened 2022-12-06
immer::map can no longer be used with std::views/std::ranges


jon-smith:
I noticed since `champ_iterator() = default;` was removed, `immer::map` no longer satisfies the constraint `std::ranges::range`, which prevents usage with views, e.g. `for(const auto &v : map | std::views::values)`.

The issue is with `std::ranges::end()`, but oddly, `std::end()` compiles fine. I'm not sure if this is a difference in the standard or a compiler issue.

For reference, I'm compiling on MSVC v19.34.31935 (latest version at present time).

I'm happy to submit a PR to reintroduce the default constructor but wanted to check if this is wanted/the right way to go about it.

arximboldi:
It would be nice to have it, for sure. I'm happy for a PR to reintroduce it. Not sure now how and why it was removed.

jon-smith:
Fixed by #257

PR #257, opened 2023-05-02
Fixes and tests for C++20 Range compatibility

vector-of-bool:
Addresses #240, as well as finding a problem with `iterator_facade` not being able to model `random_access_iterator`.

Define tests on each container that it satisfies the intended range concept. (This will transitively validate the iterators as well)

- Fix: `champ_iterator()` was not `default_initializable`, which is a required for `semiregular`, which is required for iterators.
- Fix: `iterator_facade::operator[]` returned a `reference_proxy`, while `iterator_facade::operator*` returns a `ReferenceT`. The `random_access_iterator` concept requires that the operators return an identical type.

### DIFF
--- a/immer/detail/iterator_facade.hpp
+++ b/immer/detail/iterator_facade.hpp
@@ -101,7 +101,7 @@ public:
     ReferenceT operator[](DifferenceTypeT n) const
     {
         static_assert(is_random_access, "");
-        return derived() + n;
+        return *(derived() + n);
     }
 
     friend bool operator==(const DerivedT& a, const DerivedT& b)
--- a/immer/detail/iterator_facade.hpp
+++ b/immer/detail/iterator_facade.hpp
@@ -82,19 +82,6 @@ protected:
         std::is_base_of<std::bidirectional_iterator_tag,
                         IteratorCategoryT>::value;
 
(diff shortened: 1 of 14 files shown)


immer/detail/iterator_facade.hpp
vector-of-bool, line 1: I wasn't entirely happy to just blindly delete `reference_proxy`, as I don't know if it might serve some purpose that I don't yet understand. Without it, the tests compile and run fine (on my machine), so it might have been superfluous?

Upon closer inspection, now, it looks that this change isn't correct, as the `derived() + n` is not convertible to `ReferenceT`, but this isn't causing any errors yet because no tests instantiate `iterator_facade::operator[]`. Could this the return expression be changed to `return *(derived() + n)`?

vector-of-bool: I added some small tests and confirmed that my change was bad. I changed `operator[]` to return `*(derived() + n)`, which seems to work, but I cannot be certain on why it returned a proxy reference originally, so I cannot say it is "definitely better."

arximboldi (approved):

arximboldi:
Thank you so much for this contribution!

PR #258, opened 2023-05-02
Fixing annoying unused local variable warning

bobkocisko:
This is an issue that is showing up in Visual Studio 2022/clang-cl and due to the nature of reporting it clutters up the build output window.

### DIFF
--- a/immer/heap/cpp_heap.hpp
+++ b/immer/heap/cpp_heap.hpp
@@ -33,7 +33,7 @@ struct cpp_heap
      * `allocate`.  One must not use nor deallocate again a memory
      * region that once it has been deallocated.
      */
-    static void deallocate(std::size_t size, void* data)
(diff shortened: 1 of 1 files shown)


arximboldi:
Can you disable the warning on your compiler? I always found that this is kinda a stupid warning and the code actually is better with the name in the argument... Even though this is not the first time this comes up, so maybe I bite the bullet...

bobkocisko:
Interesting!  I generally prefer to leave the warning enabled because it has sometimes helped me remember to finish something I started when a variable gets left dangling.  If you prefer, I could re-submit a different fix such as `/* inline commenting */` the variable name or using the `(void)varname` trick?

khiner:
Fwiw, I have also considered submitting a PR to somehow address these unused variables to help reduce the noise in my builds.

Issue #254, opened 2023-04-05
Logo is hidden when using GitHub dark theme


tocic:
The logo is partially hidden when you look at it using the default dark theme.

"Dark default" vs "Light default" for the `README`:

![image](https://user-images.githubusercontent.com/22129543/230119963-b4caa68a-4cf4-40d5-9a3c-1c41a4f4fa81.png)

![image](https://user-images.githubusercontent.com/22129543/230120408-880a8bbc-0606-4bed-9f24-8a9f4c37997d.png)

Consider changing the background to white.

arximboldi:
good catch!

PR #259, opened 2023-05-08
Use alternative logo in dark mode

tocic:
I've noticed that you already have an alternative version of the logo, so we can differentiate them according to the current theme.

Closes https://github.com/arximboldi/immer/issues/254.


arximboldi:
Thank you!

PR #263, opened 2023-05-24
Use destroy_at to call explicit destructor in make_inner_n

colugomusic:
This fixes the MSVC madness I was experiencing before. This compiles in both C++17 and C++20 mode.


arximboldi (approved):
Thanks for fixing this! :heart:

PR #266, opened 2023-07-29
Switch away from doctest, and drop it

pinotree:
There is only one test using doctest instead of catch2; hence, switch it to catch2, so doctest can be dropped.

This is also prompted by the fact that the old doctest copy does not build with glibc 2.34 (released 2 years ago [1]), because `SIGSTKSZ` is no more a constant; this was fixed long ago in doctest upstream [2][3]. Since we can get away from doctest, let's get rid of it rather than patching/updating an old copy.

[1] https://sourceware.org/pipermail/libc-alpha/2021-August/129718.html
[2] https://github.com/doctest/doctest/issues/473
[3] https://github.com/doctest/doctest/commit/099d5414e97244ec44cf46b14cd176b3a3dc52e3


arximboldi:
Thank you very much!

PR #267, opened 2023-07-29
Use Catch2 as installed

pinotree:
When building the tests, require and use Catch2 as installed in the system, rather than using an old embedded copy of it. This way, there is no risk it will stop working with newer versions of glibc/gcc/etc.

This also changes the nix packaging to disable tests & examples: they are not used in any way, so better disable them to avoid requiring bits only needed for them.


arximboldi:
Dam, two of the jobs seem to be failing! (the last one about benchmarks can be ignored).

pinotree:
The CIFuzz issue is due to the oss-fuzz configuration for immer that needs to be updated -- see https://github.com/google/oss-fuzz/pull/10759.

arximboldi:
Will merge once https://github.com/google/oss-fuzz/pull/10759 is merged!

pinotree:
> Will merge once [google/oss-fuzz#10759](https://github.com/google/oss-fuzz/pull/10759) is merged!

It should be good now? :)

arximboldi:
It is, great!

arximboldi:
Thank you so much @pinotree for taking care of updating this in all the repos!

PR #268, opened 2023-08-16
cmake: drop FindGuile.cmake

pinotree:
It seems it is actually not used, as guile is currently searched with

    pkg_check_modules(Guile guile-2.2)

since commit 026f2fe4bf7c4a2b62b670d76f6488929e1cded6 -- hence, drop this module.


arximboldi:
Great!

PR #269, opened 2023-08-19
cmake: use ${CMAKE_CURRENT_SOURCE_DIR} where possible

pinotree:
This makes things rely less on their current relative locations in the sources.


arximboldi:
Nice!

Issue #171, opened 2021-03-02
Implement for_each_chunk_p for champ


arximboldi:
This would fix `all_of` for maps and sets.

PR #270, opened 2023-09-23
champ for_each_chunk_p

fabianbs96:
Currently, the immer `set`, `map`, and `table` containers only support a subset of available algorithms; especially `immer::all_of` is not supported. That is, because the underlying `champ` does not implement `for_each_chunk_p`.

This PR adds an implementation of `for_each_chunk_p` to the above mentioned `champ`. Should be related to #171.

Design considerations:
- iterative- vs recursive implementation: As `for_each_chunk_p may be part of extremely hot parts of user-code, I prefer to avoid non-tail recursion
- Worklist vs explicit stack: The easiest non-recursive implementation would be based on a worklist of `const node_t *`. However, this has the drawback of potential memory allocation, so the implementation uses an explicit stack (`std::array`) that models the required parts of the call-stack from `for_each_chunk_traversal`. 
- Question: Should we `std::invoke` the callback if compiled with C++17 or higher?

### DIFF
--- a/immer/detail/hamts/champ.hpp
+++ b/immer/detail/hamts/champ.hpp
@@ -12,6 +12,8 @@
 #include <immer/detail/hamts/node.hpp>
 
 #include <algorithm>
+#include <array>
+#include <cassert>
 
 namespace immer {
 namespace detail {
@@ -146,7 +148,8 @@ struct champ
     champ(node_t* r, size_t sz = 0)
         : root{r}
         , size{sz}
-    {}
+    {
+    }
 
     champ(const champ& other)
         : champ{other.root, other.size}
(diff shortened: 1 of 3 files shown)


arximboldi (commented):
This is a great contribution thank you! Once gain, sorry for the delays in reviewing this...

...one of the reasons it took me long to review it, is I wanted to book time to understand the implementation properly, as a recursive implementation would have been easier to understand. Have you actually tried to benchmark this and see if there is an actual performance benefit and, if so how much? It feels to me that you're doing more or less the same work than the compiler generates when using recursion normally (just a bit less, not saving the pointer to the code position). But I ask out of genuine curiosity as I really can't predict here what performance would be like, as modern compilers and processors are sometimes surprising when it comes to micro-optimizing...

fabianbs96:
Hi @arximboldi,
when I implemented this, I also did some benchmarking, but I don't remeber the numbers anymore. You are right, modern compilers and processors already do a lot of optimizations, but in my experience (since I work in the area of static code analysis based on LLVM) compilers are especially bad, when it comes to recursion, so I try to avoid it whenver I can.
For example, the recursive calls are usually not inlined, although in this particular situation, inlining would make a lot of sense.

I might do some benchmarks to show some numbers, but I currently cannot tell you when I will find the time to do this.

PR #275, opened 2024-01-16
Fix portability of IMMER_THROW when exceptions are disabled

tjaneczko:
The current implementation of IMMER_THROW when exceptions are disabled relies on the compiler converting a character array to a truthy value, which can throw warnings and is not truly portable. Instead, we can use the comma operator to portably add additional data to an assertion: https://en.cppreference.com/w/cpp/error/assert.


arximboldi:
Nice one!

PR #276, opened 2024-01-17
cmake: Only use ccache for C++

patstew:
Setting RULE_LAUNCH_* breaks the build of anything other than C/C++ that ccache doesn't support, e.g. rc.exe on windows. The global property leaks out and breaks any consuming project.


arximboldi:
Make sense, thanks!

PR #278, opened 2024-02-14
Immer persist

alex-sparus:
Structural sharing is traditionally preserved in memory only. When serializing multiple values to disk, these are linearized—for example, by writing out sequences as JSON arrays—losing whatever sharing may exist between these values. A similar thing happens when transforming data structures (as with `std::transform` or a functional `map()`)

Here we introduce an experimental module `immer::persist` that enables the serialization of pools of containers while preserving structural sharing. With it, we can also do in-memory transformation over large sets of data that considers and preserves structural sharing. This is novel technique, not present in any other implementation of persistent data structures. It has a few use-cases inside BRONZE.


immer/extra/persist/README.rst
```diff
@@ -79,8 +79,22 @@ Which generates some JSON like this:
 As you can see, the value is serialized with every ``immer`` container replaced by an identifier.
 This identifier is a key into a pool, which is serialized just after.
 
-- XXX: what's a pool?
-- XXX: to_json_with_pool uses cereal
+A pool is a data structure that ``immer-persist`` uses to work with the underlying structure of the ``immer`` containers.
+For example, ``vector`` and ``flex_vector`` use Radix Balanced Trees internally that have inner nodes and leaf nodes. These
+nodes can be seen in the JSON above. In other words, containers can be put into a pool and later can be retrieved from it, all while preserving
+the structural sharing. One pool can only work with one container type.
```
arximboldi, line 85: Try to think from the user's point of view, instead of from the implementors point of view. For example, here you say that the pool is a data-structure that is used to work with the underlying structure and then go to talk about nodes. That is kind of true for the implementer, but the user doesn't care so much about nodes and what not.

This is an alternative explanation, that focuses on, from the reader's POV, what the pool is conceptually (what a pool value represents) and then why it's needed (which looks at the implementation but from the POV of why this is useful to the user).

> A pool represents a _set_ of Immer containers of a given type. For example, we may have a pool that contains all `immer::vector<int>` of our document. You can think of it as a little data base of Immer containers. When serializing the pool, the internal structure of all those Immer containers is written, preserving the structural sharing between those containers. The nodes of the trees that implement the Immer containers are represented directly in the JSON and, because we have are representing all the containers as a whole, those nodes that are referenced in multiple trees can be stored only once. That same structure is preserved when reading the pool back from disk and reconstructing the vectors from it, thus allowing us to preserve the structural sharing across sessions.

arximboldi: Also maybe worth mentioning the distinction between output and input pool. Unless there is a away to make it so that only one type of pool exists in the code, without much performance penalty? That would make it simpler for the user.

arximboldi (changes requested):
This is looking very good, I think this is almost done. Do you think this is ready to get the `draft` marker gone?

A couple of polishing remarks, mostly cosmetic stuff, some questions, etc. I'm sorry I'm not doing inline comments, but the diff is so big that the Github UI goes crazy when I try to do so.

1. I see you've introduced also `immer/extra/cereal/...` with some of the serializing code we use in so many other projects. I think this is good, but should we make this complete and include serielizers for all immer types, and get rid of this in the Lager repo: https://github.com/arximboldi/lager/tree/master/lager/extra/cereal

2. I suggest splitting the file `immer/extra/persist/cereal/with_pools.hpp` into `immer/extra/persist/cereal/save.hpp` and `immer/extra/persist/cereal/load.hpp`. These seem to be the main files the library is providing and it is not so clear what the file is about from looking at the current name.

3. When looking at the pool internal interface, there is a confusing mix of free functions and methods. For example, `save` is a method of `output_pool` but, but `add_to_pool` is not. Maybe they should both be methods? Also what about adding a bit encapsulation? I.e. making the data, and methods that are not meant to be used externally, private?

4. It seems `detail/common/pool.hpp` is dead code. Can it be removed? Are there any other bits of code that I'm missing that can be removed?

5. It seems some of the implementation files introduce a depednency on `spdlog`, mostly for debugging/trace messages. Can we get rid of it?

6. The documentation should mention early that this library introduces dependencies additional to `immer`: C++17 (or even C++20?), `boost::hana` (other parts of boost as well?) and for the most part `cereal` also. 

7. There are other changes I'd like to make to the docs, but I think I will do that once we're merged as it is easier than me making changes into your repo... Please remind me to give you write acces into the `immer` repo, so future PR's you can make directly from a branch in `immer` :slightly_smiling_face: 

8. Talking about the documentation: it doesn't seem to indicate why the types need to be part of `boost::hana` by default. It could be an interesting thing to mention. Also I personally think the example code would be clearer if `BOOST_HANA_DEFINE_STRUCT` wasn't used, but instead we used the alternative syntax where the struct is defined normally and then separately the struct is lifted into Hana (it requires a few more characters, but the code is clearer in my opinion, particularly for people not familiar with Hana, as one doesn't need to understand the internals of Hana in order to know that the macro expands to mostly just a normal struct).

Overall excellent work! I think this is a massive use-case, it opens so many possibilities. And a unique feature not provided by any other immutable data-structures library I know of. Something worth writing a paper about probably! Looking forward to see this get more use and to evolve, so we can eventually treat it as a first-class citizen of `immer`.

arximboldi:
Alright! Thank you for all this work!

PR #281, opened 2024-03-12
Fixes `difference_type` being mispelled as `diference_type`.

nicola-gigante:
This fixes the type alias `difference_type` which was mispelled as `diference_type` in `map`, `set`, `table`, and their transients. This may potentially break some generic code, although these aliases are not used as much as in the past nowadays.


arximboldi (approved):
Good catch, thank you!

nicola-gigante:
Thanks to you for this wonderful library!

PR #284, opened 2024-04-28
Remove usage of `std::aligned_storage` to switch off deprecation warnings in C++23

### LINKED ISSUE #283: Compiling in C++23 mode issues a deprecation warning for `std::aligned_storage`.
First of all thanks for the awesome library!

The new LLVM/Clang 18 implemented the deprecation of `std::aligned_storage` in C++23 and therefore it warns the following when compiling a project using `immer` in C++23 mode:

```
.../immer/detail/util.hpp:40:1: error: 'aligned_storage<8, 8>' is deprecated [-Werror,-Wdeprecated-declarations]
   40 | using aligned_storage_for =
```

of course one can switch off the warning but ideally this would be avoided by stopping using `std::aligned_storage`. 

There are some ways to avoid it as written here: https://stackoverflow.com/questions/71828288

The replacement in the SO answer cites `std::byte`, which came in C++17 afaik, but if `immer` wants to support C++14 it should be ok to just use `char`, so this means the line above would become something like:

```
template<typename T>
struct aligned_storage_for {
    alignas(T) char data[sizeof(T)] = { };
};
```

I can prepare a PR if you are interested in this fix.

nicola-gigante:
This PR effectively adds a simple implementation of `std::aligned_storage` in the `immer::detail` namespace to avoid using the `std` one which causes deprecation warnings. This is meant to close #283.

### DIFF
--- a/immer/detail/util.hpp
+++ b/immer/detail/util.hpp
@@ -36,9 +36,20 @@ const T& as_const(T& x)
     return x;
 }
 
+template<std::size_t Size, std::size_t Align>
+struct aligned_storage
+{
+    struct type {
+        alignas(Align) unsigned char data[Size] = { };
+    };
+};
+
+template<std::size_t Size, std::size_t Align>
+using aligned_storage_t = typename aligned_storage<Size, Align>::type;
+
 template <typename T>
 using aligned_storage_for =
-    typename std::aligned_storage<sizeof(T), alignof(T)>::type;
+    typename aligned_storage<sizeof(T), alignof(T)>::type;
 
 template <typename T>
 T& auto_const_cast(const T& x)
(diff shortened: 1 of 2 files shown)


nicola-gigante:
Hi! As you see some tests do not pass. I actually have some problems running them locally. I do not have Nix and as I work on macOS it's not very pleasant to install. Can I run the tests without Nix?

nicola-gigante:
Ok, now tests seem to pass except one that seems unrelated to the PR itself.

maierlars (approved):

arximboldi (approved):
Thank you a lot @nicola-gigante ! Merging :)

nicola-gigante:
You are welcome!

PR #285, opened 2024-07-23
set/map nothrow move constructible/assignable

jj683:
Based entirely on @maierlars work , done in #244.


maierlars (approved):

arximboldi (approved):
This is great, thank you!

PR #289, opened 2024-09-11
Make persist compatible with C++17

alex-sparus:


arximboldi (approved):
This is pretty nice, thank you!

PR #292, opened 2024-09-18
Support hash size that's not std::size_t

alex-sparus:
TODO:
- Make a test with WASM and try to read a map from JSON


arximboldi (approved):
LGTM!

PR #295, opened 2024-11-13
Mark package as arch independent

sertonix:
Otherwise the generated file will check whenever the 32/64 bit-ness matches. Since this is a header only library this check shouldn't exist.

Ref https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#generating-a-package-version-file


arximboldi:
Thank you very much!

It seems there are some formatting issues in the CMake code, can you please run `cmake-format`, ideally in the `nix-shell` so it uses the same version as we do... also running the `nix-shell` will set up some pre-commit hooks that will prevent this issue in the future.

sertonix:
This seems to make cmake-format pass but it is basically just working around the fact that cmake-format isn't aware of `ARCH_INDEPENDENT`: [cmakelang/parse/funs/standard_modules.py#L1891](https://github.com/cheshirekow/cmake_format/blob/eff5df1f41c665ea7cac799396042e4f406ef09a/cmakelang/parse/funs/standard_modules.py#L1891)

arximboldi:
Thank you! Agree that it looked better before but tooling is key :)

PR #297, opened 2025-03-13
Add default ctor ability to gc policy's edit

jeaye:
This code was incorrect before, from everything I can tell. I suspect a Clang issue for why it never failed before. It boils down to this:

* The GC policy's `edit` type is not default constructible: https://github.com/arximboldi/immer/blob/master/immer/transience/gc_transience_policy.hpp#L47
* map, set, and vector all default construct their `edit` in a few places

What I did to make it show up:

* I moved my immer map, vector, set instantiations to an `extern template`, with one definition in a `.cpp` file

Originally, I fixed this by changing `{}` to `{nullptr}`, but that won't work for other policies, such as this: https://github.com/arximboldi/immer/blob/master/immer/transience/no_transience_policy.hpp#L24

Instead, I have made the GC policy's `edit` default constructible. The `void*` ctor was removed, since aggregate initialization can now just be used instead.

### DIFF
--- a/immer/transience/gc_transience_policy.hpp
+++ b/immer/transience/gc_transience_policy.hpp
@@ -39,12 +39,8 @@ struct gc_transience_policy
 
             struct edit
             {
-                void* v;
-                edit(void* v_)
-                    : v{v_}
-                {
-                }
-                edit() = delete;
+                void* v{};
+
                 bool operator==(edit x) const { return v == x.v; }
                 bool operator!=(edit x) const { return v != x.v; }
             };


jeaye:
@arximboldi Any interest in this fix?

arximboldi:
Hi @jeaye ! I am interested in this. I just wanted to find some time to study this commit properly to see if there are any potential unwanted cosequences.

I also wanted to look deeper into this, which I've wondered if it is related. I wonder whether there are current bugs when using `gc_transience_policy`, as it is less common in the production systems I've seen. https://github.com/NixOS/nix/pull/11767#issuecomment-2512295735

jeaye:
> I also wanted to look deeper into this, which I've wondered if it is related. I wonder whether there are current bugs when using `gc_transience_policy`, as it is less common in the production systems I've seen. [NixOS/nix#11767 (comment)](https://github.com/NixOS/nix/pull/11767#issuecomment-2512295735)

Ah, this would be very good to know. Please let me know what you discover, or if you need any help testing. There's no rush to merge this for jank, since we're running on our fork of immer. Just wanted to make sure this had some momentum and I see now that it does. :slightly_smiling_face: 

Thank you, Juanpe!

arximboldi:
Sorry it took so long. Finally merged!

jeaye:
> Sorry it took so long. Finally merged!

Thanks! I forgot all about this. :joy:

Issue #302, opened 2025-07-25
std::vector creation from immer::map/set iterator pair fails starting at gcc 14.3


TheCoconutChef:
The following fails to compile starting from gcc 14.3 ([godbolt link](https://godbolt.org/z/b7brMsdrx)):

```c++
#include <immer/map.hpp>
#include <unordered_map>
#include <vector>

int main()
{
    auto um = std::unordered_map<int,int>{ {0,0}, {1,1}, {2,2} };
    auto uv = std::vector<std::pair<int,int>>(um.begin(), um.end());
    static_assert(!std::sized_sentinel_for<decltype(um.end()), decltype(um.begin())>);

    auto im = immer::map<int,int>{ {0,0}, {1,1}, {2,2} };
    auto iv = std::vector<std::pair<int,int>>(im.begin(), im.end());
    static_assert(std::sized_sentinel_for<decltype(im.end()), decltype(im.begin())>);

    return 0; 
}
```

with the following:

```
/opt/compiler-explorer/gcc-14.3.0/include/c++/14.3.0/bits/ranges_base.h:961:23:   required from 'constexpr std::iter_difference_t<typename std::decay<_Tp>::type> std::ranges::__distance_fn::operator()(_It&&, _Sent) const [with _It = immer::detail::hamts::champ_iterator<std::pair<int, int>, immer::map<int, int>::hash_key, immer::map<int, int>::equal_key, immer::memory_policy<immer::free_list_heap_policy<immer::cpp_heap>, immer::refcount_policy, immer::spinlock_policy>, 5>&; _Sent = immer::detail::hamts::champ_iterator<std::pair<int, int>, immer::map<int, int>::hash_key, immer::map<int, int>::equal_key, immer::memory_policy<immer::free_list_heap_policy<immer::cpp_heap>, immer::refcount_policy, immer::spinlock_policy>, 5>; std::iter_difference_t<typename std::decay<_Tp>::type> = long int; typename std::decay<_Tp>::type = immer::detail::hamts::champ_iterator<std::pair<int, int>, immer::map<int, int>::hash_key, immer::map<int, int>::equal_key, immer::memory_policy<immer::free_list_heap_policy<immer::cpp_heap>, immer::refcount_policy, immer::spinlock_policy>, 5>]'
  961 |       { return __last - static_cast<const decay_t<_It>&>(__first); }
[...]

PR #303, opened 2025-08-20
Use sfinea to conditionally exclude iterator facade operators

TheCoconutChef:
This fixes an issue that showed up on recent compiler version -- e.g. gcc 14.3 -- in which the std::sized_sentinel_for concept would evaluate to true on forward iterator types in spite of the fact that their operator- would fail to compile. This only occured with C++20 or greater. This prevented doing things like

  auto m = immer::map<int,int>{{0,0}, {1,1}}
  auto v = std::vector<std::pair<int,int>>(m.begin(), m.end());

or

  auto n = std::ranges::distance(m.begin(), m.end());

This was due to the fact that static_asserts were used to exclude certain operators at compile time, but C++ concepts basically ignore those. This would then lead the standard library to chose implementations relying on operators that couldn't compile.

Link:

https://en.cppreference.com/w/cpp/iterator/sized_sentinel_for.html

Issue:

https://github.com/arximboldi/immer/issues/302

### DIFF
--- a/immer/detail/iterator_facade.hpp
+++ b/immer/detail/iterator_facade.hpp
@@ -98,9 +98,11 @@ protected:
 public:
     ReferenceT operator*() const { return access_t::dereference(derived()); }
     PointerT operator->() const { return &access_t::dereference(derived()); }
+
+    template <typename U = iterator_facade,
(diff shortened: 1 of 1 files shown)


arximboldi:
Delecious coconut! Thanks a lot Chef! :+1:

PR #315, opened 2025-12-21
Fix configure on FreeBSD

BSDKaffee:
Include correct CMake module for check_c_source_runs() in [cmake/FindBoehmGC.cmake](https://github.com/arximboldi/immer/blob/master/cmake/FindBoehmGC.cmake).

On FreeBSD, configure fails with:
```
CMake Error at cmake/FindBoehmGC.cmake:67 (check_c_source_runs):
  Unknown CMake command "check_c_source_runs".
```
I'm the maintainer for the [devel/immer](https://cgit.freebsd.org/ports/tree/devel/immer) port on FreeBSD and have been patching this locally for over 2 years, so this change has been pretty well tested.


arximboldi:
This LGTM, thank you!

PR #325, opened 2026-02-25
CMake: improvements

timblechmann:
* by default enable tests/examples/docs only if the project is the top-level project
* use CMAKE_XXX_LAUNCHER variables, as the RULE_LAUNCH properties are internal to cmake
* add alias target, to make linking more robust
* expose headers to IDEs via FILE_SET

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.21)
 cmake_policy(SET CMP0048 NEW) # enable project VERSION
 cmake_policy(SET CMP0056 NEW) # honor link flags in try_compile()
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -48,11 +48,11 @@ option(ENABLE_PYTHON "enable building python module" off)
 option(ENABLE_GUILE "enable building guile module" off)
 option(ENABLE_BOOST_COROUTINE "run benchmarks with boost coroutine" off)
 
-option(immer_BUILD_TESTS "Build tests" ON)
+option(immer_BUILD_TESTS "Build tests" "${PROJECT_IS_TOP_LEVEL}")
```
timblechmann, line 51: PROJECT_IS_TOP_LEVEL was introduced in 3.12. it could be emulated via `CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME`, but i wonder if that's worth the effort: cmake 3.21 is from 2021, so widely available already

CMakeLists.txt
```diff
@@ -138,7 +139,111 @@ if(immer_DISABLE_EXCEPTIONS)
   target_compile_definitions(immer INTERFACE IMMER_NO_EXCEPTIONS)
 endif()
 
-install(TARGETS immer EXPORT ImmerConfig)
+target_sources(immer PUBLIC FILE_SET HEADERS
+  FILES
```
timblechmann, line 143: exposing the headers to IDE project files ... quite a bit of a quality-of-life improvement

CMakeLists.txt
arximboldi: Sounds good :+1:

CMakeLists.txt
arximboldi: What happens if I forget to update it?

arximboldi:
Looking good, thank you! Will merge this and the ones in `zug` and `lager` once CI is green.

arximboldi (approved):

CMakeLists.txt
timblechmann: afaict it is mainly cosmetic, i.e. it will allow IDEs to populate targets with headers and aid indexers (at least that's how it's used with qtcreator/clangd). if a file is missing, it won't show up in IDEs, but afaict that's it ...

